PR #5, opened 2015-02-10
change look of the setup.py file

Midnighter:
* add shebang so that it can be made executable
* use __future__ for Python 2&3 compatibility
* condense imports

Besides the from `__future__` imports this doesn't change anything of significance. This is just my preferred way of organization and in case you like, you can add it.


ionelmc:
Parts of this are in 6ba703c110e74020019a0f9b4fc61f4e7acceb10

PR #6, opened 2015-02-10
change path logic

Midnighter:
* improve logic of using paths
* modify imports
* unify use of `"` quotation instead of `'`

This pull request simplifies the logic of the paths in the `bootstrap.py` script. Importantly, it also reads the `activate_this.py` as binary. The rest is personal preference.

### DIFF
--- a/{{cookiecutter.repo_name}}/ci/bootstrap.py
+++ b/{{cookiecutter.repo_name}}/ci/bootstrap.py
@@ -5,6 +5,7 @@
 from __future__ import (absolute_import, unicode_literals, print_function)
 
 
+import os
 import sys
 from os.path import (join, exists)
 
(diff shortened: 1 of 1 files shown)


ionelmc:
Thanks for this, looks much cleaner.

One question tho, how come you moved all the imports on a single line? I usually have them on individual as to easily see which imports are unused in my editor (the linter would underline the whole line).

Midnighter:
I try not to waste too much vertical space and thus prefer this way of importing. I code in gvim with syntastic and flake8 support and it will tell me exactly which of those imports are missing/redundant, i.e., underline only the specific import and not the whole line. It comes down to what you prefer really.

ionelmc:
Rebased, seems I caused some conflicts.

I hope you don't mind the amendments, I'm too old to learn vim :grimacing:

Midnighter:
Haha, fair enough. Whatever works for you. Also, feel free to just trash #5.

ionelmc:
Allright, thanks.

PR #9, opened 2015-03-23
Add `test_runner` option with two possible values: pytest and nosetests.

svetlyak40wt:


ionelmc:
You need to add `nose` to the dependencies (and the coverage plugin maybe?).

svetlyak40wt:
Good catch! Forgot them.

ionelmc:
I think I forgot to mention the help section in the readme: https://github.com/ionelmc/cookiecutter-pylibrary/blob/master/README.rst#usage - there should be row about the `test_runner` variable.

One more thing: I wonder if the option name for `nosetests` shouldn't be just `nose`. What do you think?

ionelmc:
Look good. I've pushed it to the develop branch. Gonna push it to master after some testing, seems appveyor is useless.

svetlyak40wt:
Ok, thank you!

PR #11, opened 2015-06-04
Update tox.ini for coveralls compatibility with tox 2.0

cmeeren:
See coagulant/coveralls-python#73 and referenced issues/PRs. This fixed the issue for me.

### DIFF
--- a/{{cookiecutter.repo_name}}/ci/templates/tox.ini
+++ b/{{cookiecutter.repo_name}}/ci/templates/tox.ini
@@ -13,6 +13,7 @@ envlist =
 setenv =
     PYTHONPATH={toxinidir}/tests
     PYTHONUNBUFFERED=yes
+passenv = *
 deps =
 {%- endraw -%}
 {%- if cookiecutter.test_runner|lower == "pytest" %}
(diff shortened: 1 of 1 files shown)


ionelmc:
Good catch!

I'm also considering having a `passenv = *` - there are many other cases [1] that break because of missing env vars.

[1] 
  https://bitbucket.org/hpk42/tox/issue/248/pass-through-vs-comntools-during-install
  https://bitbucket.org/hpk42/tox/issue/254/tox-2-fails-with-unicodedecodeerror-trying
  https://bitbucket.org/hpk42/tox/issue/253/newer-tox-prevents-usage-of-ssh-agent-201

cmeeren:
Good idea, I don't see any reason not to do that (that's how it used to work, after all). Will that need to be added in each ``[testenv:foo]`` or is it in effect for all the environments if it is added to ``[testenv]``?

ionelmc:
Adding it in `[testenv]` is sufficient.

cmeeren:
Ok. Will you do that yourself or should I modify my PR? Or haven't you decided yet?

ionelmc:
You can make the change if you want. I'll all some info in the readme about why we allow all vars after that.

ionelmc:
Alright, thanks.

PR #19, opened 2015-06-09
Use readthedocs sphinx theme

cmeeren:
Since you already have the readthedocs badge in the readme, I assume it is desirable to use the readthedocs theme locally too see how it will turn out before pushing. The whole ``on_rtd`` business is [a documented fix](https://docs.readthedocs.org/en/latest/theme.html#how-do-i-use-this-locally-and-on-read-the-docs).

### DIFF
--- a/{{cookiecutter.repo_name}}/docs/conf.py
+++ b/{{cookiecutter.repo_name}}/docs/conf.py
@@ -26,10 +26,6 @@ author = {{ '{0!r}'.format(cookiecutter.full_name) }}
 copyright = '{0}, {1}'.format(year, author)
 version = release = {{ '{0!r}'.format(cookiecutter.version) }}
 
-import sphinx_py3doc_enhanced_theme
-html_theme = "sphinx_py3doc_enhanced_theme"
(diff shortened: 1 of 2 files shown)


ionelmc:
The theme in there is something custom, not the RTD theme. It looks like http://python-aspectlib.readthedocs.org/en/latest/

I can see that there's appeal to just use RTD's theme by default. Maybe we should have some comments around for instructions on how to use a custom theme?

cmeeren:
That, or a question "use RTD doc theme?" when running cookiecutter.

ionelmc:
Yes, that would be better :-)

cmeeren:
I'll leave that up to you then, I have no knowledge of cookiecutter's internals or jinja :)

ionelmc:
Actually it's not that hard, eg: https://github.com/ionelmc/cookiecutter-pylibrary/commit/ddcb4136319e6f38e60aaa4f23231d5e3e1d6048

ionelmc:
Give it a shot :grin:

PR #20, opened 2015-06-09
Add option to use sphinx readthedocs theme locally

cmeeren:


ionelmc:
Alright, in https://github.com/ionelmc/cookiecutter-pylibrary/commit/17ca2dcda9e071a73ed4ebb0ebd3ff38580c68c0

However, I've changed it a bit so you can use any theme.

PR #22, opened 2015-07-21
Adding Codacy and CodeClimate badges

kaidokert:
Added, here is a quick test of what they look like : https://github.com/kaidokert/python-pytest1


ionelmc:
Thanks!

PR #35, opened 2015-11-12
Improve .travis.yml build matrix

abusalimov:
First and foremost, thank you for the comprehensive template.

This PR improves the way a build matrix is constructed for different Python versions. Instead of having just a single `TOXENV` axis defining a 1D matrix like this:
```yaml
env:
  matrix:
    - TOXENV=check

    - TOXENV=2.6-cover,codecov
    - TOXENV=2.6-nocov

    - TOXENV=2.7-cover,codecov
    - TOXENV=2.7-nocov

    ...
```

... use distinct Python versions as an additional axis, in a natural way for Travis CI:
```yaml
python:
  - "2.6"
  - "2.7"
  ...
env:
  matrix:
    - COVERAGE=cover,codecov
    - COVERAGE=nocov
```

... and infer `TOXENV` from that:
```yaml
before_script:
  - export TOXENV=${TOXENV:-$TRAVIS_PYTHON_VERSION-$COVERAGE}
```

This makes the Travis UI look better since true Python versions are shown for each build row. A similar change has been merged into a pytest-plugin cookiecutter template recently: pytest-dev/cookiecutter-pytest-plugin#13

Also make few less important changes, mostly cleanup.

NB: The last commit adds installation of [pytest-travis-fold](https://github.com/abusalimov/pytest-travis-fold), which in turn may be considered as self-advertising (I'm the author of the plugin). If you believe this is inappropriate, please let me know, and I will drop that commit.

### DIFF
--- a/{{cookiecutter.repo_name}}/.travis.yml
+++ b/{{cookiecutter.repo_name}}/.travis.yml
@@ -15,14 +15,10 @@ env:
     - TOXENV={{ env }}-nocov
 {% endfor %}
 before_install:
-  - python --version
-  - uname -a
-  - lsb_release -a
+  - easy_install --version
 install:
   - pip install tox
   - virtualenv --version
-  - easy_install --version
-  - pip --version
   - tox --version
 script:
   - tox -v
--- a/{{cookiecutter.repo_name}}/.travis.yml
+++ b/{{cookiecutter.repo_name}}/.travis.yml
@@ -1,6 +1,9 @@
-language: python
-python: '3.5'
 sudo: false
+language: python
+python:
+{% for env in ['2.6', '2.7', '3.3', '3.4', '3.5', 'pypy'] %}
+  - "{{ env }}"
+{% endfor %}
 env:
 {% if cookiecutter.c_extension_support|lower == "yes" %}
   global:
@@ -8,28 +11,33 @@ env:
     - SEGFAULT_SIGNALS="abrt segv"
 {% endif %}
   matrix:
-    - TOXENV=check
-{% for env in ['2.6', '2.7', '3.3', '3.4', '3.5', 'pypy'] %}
-    - TOXENV={{ env }}-cover
+    - COVERAGE=cover
       {%- if cookiecutter.coveralls|lower == 'yes' %}
         {%- if cookiecutter.c_extension_support|lower == "yes" %},extension-coveralls{% endif %},coveralls
       {%- endif %}
       {%- if cookiecutter.codecov|lower == 'yes' %},codecov{% endif %}
-    - TOXENV={{ env }}-nocov
-{% endfor %}
+    - COVERAGE=nocov
+matrix:
+  include:
+    - python: "3.4"
+      env: TOXENV=check
+
 before_install:
   - easy_install --version
 install:
   - pip install tox
   - virtualenv --version
   - tox --version
+
+before_script:
+  - export TOXENV=${TOXENV:-$TRAVIS_PYTHON_VERSION-$COVERAGE}
 script:
   - tox -v
 after_failure:
   - more .tox/log/* | cat
   - more .tox/*/log/* | cat
+
 notifications:
   email:
     on_success: never
     on_failure: always
-
--- a/{{cookiecutter.repo_name}}/.travis.yml
+++ b/{{cookiecutter.repo_name}}/.travis.yml
@@ -26,10 +26,14 @@ before_install:
   - easy_install --version
 install:
   - pip install tox
+    {%- if cookiecutter.test_runner|lower == "pytest" %} pytest-travis-fold{% endif %}
   - virtualenv --version
   - tox --version
 
 before_script:
+{% if cookiecutter.test_runner|lower == "pytest" %}
+  - export TOX_TESTENV_PASSENV=TRAVIS
+{% endif %}
   - export TOXENV=${TOXENV:-$TRAVIS_PYTHON_VERSION-$COVERAGE}
 script:
   - tox -v
--- a/{{cookiecutter.repo_name}}/.travis.yml
+++ b/{{cookiecutter.repo_name}}/.travis.yml
@@ -37,6 +37,12 @@ after_failure:
   - more .tox/log/* | cat
   - more .tox/*/log/* | cat
 
+before_cache:
+  - rm -rf $HOME/.cache/pip/log
+cache:
+  directories:
+    - $HOME/.cache/pip
+
 notifications:
   email:
     on_success: never
--- a/{{cookiecutter.repo_name}}/.travis.yml
+++ b/{{cookiecutter.repo_name}}/.travis.yml
@@ -2,8 +2,11 @@ language: python
 python: '3.5'
 sudo: false
 env:
+{% if cookiecutter.c_extension_support|lower == "yes" %}
   global:
-    LD_PRELOAD=/lib/x86_64-linux-gnu/libSegFault.so
+    - LD_PRELOAD=/lib/x86_64-linux-gnu/libSegFault.so
+    - SEGFAULT_SIGNALS="abrt segv"
+{% endif %}
   matrix:
     - TOXENV=check
 {% for env in ['2.6', '2.7', '3.3', '3.4', '3.5', 'pypy'] %}
--- a/{{cookiecutter.repo_name}}/ci/templates/.travis.yml
+++ b/{{cookiecutter.repo_name}}/ci/templates/.travis.yml
@@ -43,6 +43,12 @@ after_failure:
   - more .tox/log/* | cat
   - more .tox/*/log/* | cat
 
+before_cache:
+  - rm -rf $HOME/.cache/pip/log
+cache:
+  directories:
+    - $HOME/.cache/pip
+
 notifications:
   email:
     on_success: never
--- a/{{cookiecutter.repo_name}}/ci/templates/.travis.yml
+++ b/{{cookiecutter.repo_name}}/ci/templates/.travis.yml
@@ -34,9 +34,14 @@ before_install:
   - easy_install --version
 install:
   - pip install tox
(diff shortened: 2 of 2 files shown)


abusalimov:
Almost forgot, should I add a line into the CHANGELOG?

ionelmc:
Hey,

Thanks for the changes, although lots of the things that are changed/removed here were done like that on purpose. I'll add comments inline - lets see if my assumptions still hold :)

{{cookiecutter.repo_name}}/.travis.yml
```diff
@@ -1,36 +1,53 @@
...
-    - TOXENV={{ env }}-nocov
-{% endfor %}
+    - COVERAGE=nocov
+matrix:
+  include:
+    - python: "3.4"
+      env: TOXENV=check
+
 before_install:
-  - python --version
```
ionelmc: The versions are intentionally shown as they are not equivalent with what's in the build log header.

They have helped me debug issues with caches and other pip install issues (pip is basically broken if you have eggs around). I have been burned too many times :)

{{cookiecutter.repo_name}}/.travis.yml
```diff
@@ -1,36 +1,53 @@
...
   - pip install tox
+    {%- if cookiecutter.test_runner|lower == "pytest" %} pytest-travis-fold{% endif %}
   - virtualenv --version
-  - easy_install --version
-  - pip --version
   - tox --version
+
+before_script:
+{% if cookiecutter.test_runner|lower == "pytest" %}
+  - export TOX_TESTENV_PASSENV=TRAVIS
```
ionelmc: How does this interact with `passenv = *` inside `tox.ini`? Who takes precedence?

abusalimov:
Not a problem, I could restore the necessary lines.

{{cookiecutter.repo_name}}/ci/templates/.travis.yml
```diff
@@ -1,9 +1,12 @@
...
-python: '3.5'
 sudo: false
+language: python
+python: "3.4"
 env:
+{% if cookiecutter.c_extension_support|lower == "yes" %}
   global:
-    LD_PRELOAD=/lib/x86_64-linux-gnu/libSegFault.so
+    - LD_PRELOAD=/lib/x86_64-linux-gnu/libSegFault.so
+    - SEGFAULT_SIGNALS="abrt segv"
```
ionelmc: Interesting, didn't knew about this env var. But shouldn't it be `SEGFAULT_SIGNALS=all`?

{{cookiecutter.repo_name}}/.travis.yml
abusalimov: The [corresponding](https://bitbucket.org/hpk42/tox/src/f6cca79ba7f6522893ab720e1a5d09ab38fd3543/tox/config.py#config.py-399) code performs roughly the following:
  1. Take default variables: `"PATH", "PIP_INDEX_URL", "LANG", "LD_LIBRARY_PATH"`
  2. Add these from `TOX_TESTENV_PASSENV`
  3. Add all existing environment variables matching `passenv` from `tox.ini`.

So everything should be fine. I added this on purpose, in case if user decides to remove `passenv = *` from `tox.ini`.

{{cookiecutter.repo_name}}/ci/templates/.travis.yml
```diff
@@ -1,9 +1,12 @@
-language: python
-python: '3.5'
```
ionelmc: Specifically set to 3.5 so that it's installed. Travis don't have 3.5 on all the vms and it fails down the road (missing interpreter). Correct me if they fixed this and I'm wrong.

ionelmc:
There's a problem with the matrix changes that really bothers me: you can't really regenerate `.travis.yml` out of the env names from tox. I mean, you could but it's so complicated and unreliable (you'd have to parse out the python version out of the env name). Is that change really that important?

IOW: you can use `ci/bootstrap.py` to regenerate `.travis.yml` even if you use `test_matrix_configurator=no`. See: https://github.com/abusalimov/cookiecutter-pylibrary/blob/201662c7bddf2099bd491a2964da1ff359fbd996/%7B%7Bcookiecutter.repo_name%7D%7D/ci/templates/.travis.yml#L24-L60

I like the other changes tho (the fold plugin, the SEGFAULT_SIGNALS thing and the pip cache thing).

{{cookiecutter.repo_name}}/.travis.yml
abusalimov: > The versions are intentionally shown as they are not equivalent with what's in the build log header.

How is this even possible? Travis show the fair output of running `python --version`. Here is the [command](https://travis-ci.org/abusalimov/test/jobs/90799267#L431) from the generated shell script that runs build commands, and here's it's [output](https://travis-ci.org/abusalimov/test/jobs/#L92-L93).

> I have been burned too many times :)

Please do not misunderstand, I really wonder, how could that happen! :)

ionelmc: It's a mess: https://github.com/pypa/pip/issues

I don't trust pip. I don't trust travis. They have bugs and I don't have handy references to give to you :|

Travis may have fixed issues I had but point is, I can't trust them to put the proper stuff in the environment, so I have to printout the versions, in case they mess it up.

{{cookiecutter.repo_name}}/ci/templates/.travis.yml
abusalimov: Honestly, in turn, I didn't know about `all`. :) However [the rest](http://stackoverflow.com/a/18947029) signals don't seem to be that common, IMHO:
  - `SIGBUS` is a HW fault, there's nothing we could do to fix it
  - `SIGSTKFLT` is "stack fault on coprocessor", unused on Linux
  - `SIGILL` and `SIGFPE` are more real-world though

What's about setting it to the explicit `SEGFAULT_SIGNALS="ill abrt fpe segv"`? Suggesting just because _explicit is better than implicit_.

ionelmc: Too much cruft. Too platform specific I guess. I like short forms.

{{cookiecutter.repo_name}}/.travis.yml
abusalimov: What I am trying to say is that adding `python --version` just runs the same command one more time:

```
travis_cmd source\ \~/virtualenv/python2.7/bin/activate --assert --echo --timing
travis_cmd python\ --version --echo
travis_cmd pip\ --version --echo

travis_fold start before_install.1
  travis_cmd python\ --version --assert --echo --timing
travis_fold end before_install.1
```

And we get the obvious [repetition](https://travis-ci.org/abusalimov/test/jobs/90952855#L92-L97):
```
$ python --version
Python 2.7.9
$ pip --version
pip 6.0.7 from /home/travis/virtualenv/python2.7.9/lib/python2.7/site-packages (python 2.7)
$ python --version
Python 2.7.9
```

I won't insist on this change though. Would you like me to rollback the whole a5d525fe8cc15739fda085e7be217d94ae450874 ("travis: Don't show --version for tools already shown by Travis")?

{{cookiecutter.repo_name}}/ci/templates/.travis.yml
abusalimov: OK, no problem.

{{cookiecutter.repo_name}}/.travis.yml
ionelmc: Yeah :grin: 

Plus it's exactly the same for AppVeyor, a bit off consistency don't hurt.

{{cookiecutter.repo_name}}/ci/templates/.travis.yml
abusalimov: I'm not quite sure I understand you. What's the intention of doing that?

When using 3.5 Travis indeed tells "3.5 is not installed; attempting download", but then it runs as usual:
```console
$ python --version
Python 3.5.0
```

{{cookiecutter.repo_name}}/.travis.yml
abusalimov: OK. What's about this part of the change:
> Also checking `easy_install --version` is more actual _before_ running
`pip install`, move it into `before_install`.

Let's move all --version checks to `before_install`, or leave everything completely untouched?

{{cookiecutter.repo_name}}/ci/templates/.travis.yml
ionelmc: If you have `python: 3.4` in `.travis.yml` and you run something with python 3.5 (like from tox) it will fail.

abusalimov: Ah, now I get it, thanks.

{{cookiecutter.repo_name}}/.travis.yml
ionelmc: Yeah, so the versions are listed two times. In case there are upgrade issues, I want to know what was installed before.

abusalimov:
> Is that change really that important?

It's not that important, it just makes the Travis UI more pretty.

OK, you've convinced me to drop these changes. Do you want me to make a separate PR or just rebase this one? And will you accept similar changes in your [cookiecutter-pylibrary-minimal](https://github.com/ionelmc/cookiecutter-pylibrary-minimal) (I discovered it just recently)?

ionelmc:
> OK, you've convinced me to drop these changes. Do you want me to make a separate PR or just rebase this one? 

Whatever is easiest for you.

> And will you accept similar changes in your cookiecutter-pylibrary-minimal (I discovered it just recently)?

I plan to remove that (this template has the "minimal" option). So no, not really :)

abusalimov:
> this template has the "minimal" options

not for the ci/bootstrap.py stuff unfortunately

PR #40, opened 2015-11-25
Move badges to the bottom of README.rst

cmeeren:
Just a suggestion. I don't like seeing lots and lots of badges at the top of the page. The only really important badges for the end user IMHO are docs (for obvious reasons) and pypi version (to direct the user to pypi if encountered elsewhere), though even pypi could be removed from the top since it's in the install instructions anyway. The rest can safely be moved to the bottom, for easy access whilst avoiding the "in your face" feeling you get from the gargantuan badge table when it's at the top.

### DIFF
--- a/{{cookiecutter.repo_name}}/README.rst
+++ b/{{cookiecutter.repo_name}}/README.rst
@@ -2,6 +2,34 @@
 {{ cookiecutter.project_name }}
 {{ "=" * cookiecutter.project_name|length }}
 
+|docs| |version|
+
+{{ cookiecutter.project_short_description|wordwrap(140) }}
+
+* Free software: BSD license
+
+Installation
+============
+
+::
+
+    pip install {{ cookiecutter.distribution_name }}
+
+Documentation
+=============
+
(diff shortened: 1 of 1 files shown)


ionelmc:
I've been thinking ... should we see any badges at all on PyPI? They are mostly "development concerns". If you're an user you don't care about no badges right?

cmeeren:
I can only answer for myself, and as an end-user I definitely don't. So I don't need to see any badges on PyPI or ReadTheDocs. I mean, "tests failing"? What does that mean, that I can't/shouldn't install the package from PyPI? Probably no worries, so why even show it? And showing code quality to an end user, I mean come on, that's just TMI.

As a developer I do care, so having them on GitHub is nice.

Concerning the docs and PyPI version badges I mentioned - they're not really needed anyway since 1) there is a link to the documentation in the readme, and 2) the readme instructs to install via pip anyway.

ionelmc:
I think we should strip out the badges in `setup.py` instead of moving them around. I'll try some stuff a in bit ...

cmeeren:
:+1:

ionelmc:
See commit.

PR #41, opened 2015-11-25
Update heading in README.rst

cmeeren:
Since README.rst is included in docs/index.rst, the heading will be shown in the ReadTheDocs menu. I don't think the project name is especially informative as a title in the menu, so I'm using "Overview" here in my packages. The package name is shown at the top on PyPI anyway.

Just a suggestion. :)


ionelmc:
I think the "Overview" heading should be remove from `docs/readme.rst` then - otherwise TOC will look like this:

* Overview
  * Overview

cmeeren:
Yep, you're right, I did that in my package but not in the PR. Done!

PR #42, opened 2015-11-26
Update napoleon config

cmeeren:
``napoleon_use_ivar=True`` will make class attributes look like the parameters section instead of looking like methods.

``napoleon_use_rtype=False`` will make the return type inline, which gives a consistent look between functions with multiple returns (where the type is always inline) and functions with single returns (where the return type is on its own by default).

Attributes ``year``, ``refh``, ``datafile`` looking like methods (default):

use_ivar_false

Attributes looking like parameters (proposed change):

use_ivar_true

Return type in its own section when one variable is returned, and inline if multiple variables are returned (default):

use_rtype_true

Return type always inline for consistency (proposed change):

use_rtype_false

### DIFF
--- a/{{cookiecutter.repo_name}}/docs/conf.py
+++ b/{{cookiecutter.repo_name}}/docs/conf.py
@@ -58,3 +58,6 @@ html_sidebars = {
    '**': ['searchbox.html', 'globaltoc.html', 'sourcelink.html'],
 }
 html_short_title = '%s-%s' % (project, version)
+
+napoleon_use_ivar=True
+napoleon_use_rtype=False


ionelmc:
How do the sources for your examples look?

cmeeren:
Short version:

````python
class Apex(object):
    '''Performs coordinate conversions, field-line mapping and base vector calculations.

    Parameters
    ==========
    date : float (decimal year) or :class:`datetime.date` or :class:`datetime.datetime`, optional
        Determines which IGRF coefficients are used in conversions. Uses current date as default.
    refh : float, optional
        Reference height in km for apex coordinates (the field lines are mapped to this height)
    datafile : str, optional
        Path to custom coefficient file

    Attributes
    ==========
    year : float
        Decimal year used for the IGRF model
    refh : float
        Reference height in km for apex coordinates
    datafile : str
        Path to coefficient file

    Notes
    =====
    The calculations use IGRF-12 with coefficients from 1900 to 2020 [1]_.

    References
    ==========

    .. [1] Thébault, E. et al. (2015), International Geomagnetic Reference
           Field: the 12th generation, Earth, Planets and Space, 67(1), 79,
           :doi:`10.1186/s40623-015-0228-9`.

    '''

    def qd2apex(self, qlat, qlon, height):
        '''Converts quasi-dipole to modified apex coordinates.

        Parameters
        ==========
        qlat : array_like
            Quasi-dipole latitude
        qlon : array_like
            Quasi-dipole longitude
        height : array_like
            Altitude in km

        Returns
        =======
        alat : ndarray or float
            Modified apex latitude
        alon : ndarray or float
            Modified apex longitude

        Raises
        ======
        ApexHeightError
            if apex height < reference height

        '''

        pass

    def mlon2mlt(self, mlon, datetime, ssheight=50*6371):
        '''Computes the magnetic local time at the specified magnetic longitude and UT.

        Parameters
        ==========
        mlon : array_like
            Magnetic longitude (apex and quasi-dipole longitude are always equal)
        datetime : :class:`datetime.datetime`
            Date and time
        ssheight : float, optional
            Altitude in km to use for converting the subsolar point from
            geographic to magnetic coordinates. A high altitude is used
            to ensure the subsolar point is mapped to high latitudes, which
            prevents the South-Atlantic Anomaly (SAA) from influencing the MLT.

        Returns
        =======
        mlt : ndarray or float
            Magnetic local time [0, 24)

        Notes
        =====
        To compute the MLT, we find the apex longitude of the subsolar point at
        the given time. Then the MLT of the given point will be computed from
        the separation in magnetic longitude from this point (1 hour = 15 degrees).

        '''

        pass

````

ionelmc:
I like this. I'm gonna look at the other PRs soonish

cmeeren:
Great! You see, I'm basically just configuring it like I want it myself and making PRs so I don't have to repeat it for future packages... :stuck_out_tongue_winking_eye:

PR #43, opened 2015-11-26
Specify "code contributions"

cmeeren:
If you mean "code contributions", I think that should be clarified here. When reading it without the clarification I get the impression that e.g. donations are welcome.

I have no idea what's up with the "detox" line, I did absolutely not change that (it happened twice, must be a GH bug).


ionelmc:
Could be anything. Reporting bugs is still a contribution but it's not really code. Why would contribution be defined so narrowly?

cmeeren:
You are of course right. Since this is under the heading "Feature suggestions" I was thinking that code was the implied type of contribution in this context.

ionelmc:
Ah, missed that.

PR #49, opened 2016-02-23
if no c extension support, dont include "extension-coveralls"

luzfcb:
if no c extension support, dont include "extension-coveralls", 
also suppose, if not use coverall, "extension-coveralls" should not be included too.


ionelmc:
Good catch. Thanks!

PR #52, opened 2016-03-02
support using unittest as a test runner

taleinat:
I'm trying to add support for using the built-in `unittest` module for running tests. Here is the beginning of my attempt.

Questions:
1. Are there additional files I should be editing?
2. Could you explain the reasoning and outcome of the differences in the test running commands, when running with coverage vs. running with coverage in the "split coverage" setting?
3. Do I need to add tests for this? If so, can you point me to similar examples of existing tests?

### DIFF
--- a/cookiecutter.json
+++ b/cookiecutter.json
@@ -30,7 +30,7 @@
     "c_extension_optional": ["no", "yes"],
     "test_matrix_configurator": ["no", "yes"],
     "test_matrix_separate_coverage": ["no", "yes"],
-    "test_runner": ["pytest", "nose"],
+    "test_runner": ["pytest", "nose", "unittest"],
     "command_line_interface": ["plain", "click", "no"],
     "command_line_interface_bin_name": "nameless",
     "coveralls": ["no", "yes"],
--- a/{{cookiecutter.repo_name}}/ci/templates/tox.ini
+++ b/{{cookiecutter.repo_name}}/ci/templates/tox.ini
@@ -32,6 +32,8 @@ commands =
 {%- endif %}
 {%- if cookiecutter.test_runner|lower == "pytest" %}
     {posargs:py.test -vv --ignore=src}
+{%- elif cookiecutter.test_runner|lower == "unittest" %}
+    {posargs:python -m unittest discover -v tests}
 {%- else %}
     {posargs:nosetests -v tests}
 {%- endif %}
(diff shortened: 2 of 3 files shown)


taleinat:
Figured out myself that I need to add a test running command in `ci/setup.cfg`.

I could still use some explanation about what is actually tested.

ionelmc:
Hey, thanks for the changes. Nose and unittests are sorta dead to me but I'm willing to accept the changes if you really want this. I'm still curious as to why you don't want pytest.

This template has some test support, you'd need to add an unittest entry in [ci/setup.cfg](https://github.com/ionelmc/cookiecutter-pylibrary/blob/master/ci/setup.cfg#L10) (same as for nose) and then run `ci/bootstrap.py` to regenerate some of the test configuration.

ionelmc:
Ah, one more thing: if you wanna run stuff locally it will take long time, and it will run test that are not relevant. Just update that test config and and push the change, Travis will run it.

taleinat:
Why use `unittest` to run tests instead of `pytest`:
In several projects I use a hierarchy of TestCase classes to test similar functions/classes. I tried to get the tests working properly with `pytest`, but couldn't find a good way to get it to ignore the "abstract" base classes without ignoring their "concrete" sub-classes. With `unittest discover` this is easy: Only the concrete classes mix-in `unittest.TestCase`.

taleinat:
If you plan to remove support for `nose` in the future, that would allow greatly simplifying this cookiecutter template. In that case it would probably be better to avoid accepting this pull request. Please tell me if this is your intention so that I know whether to continue to pursue this.

ionelmc:
As long as the initial goals of the template (give working configuration for ci and coverage measurements) are fulfilled I'm fine with supporting other test tools.

For reference, nose support was added in https://github.com/ionelmc/cookiecutter-pylibrary/pull/9 by @svetlyak40wt. 

I don't think replacing nose with unittest would simplify much. Plus I don't see any harm having a 3rd choice if the default choices aren't changed.

ionelmc:
@taleinat can you update the PR to include tests for the unitest config (as indicated in https://github.com/ionelmc/cookiecutter-pylibrary/pull/52#issuecomment-191408006)?

ionelmc:
@taleinat hey, I can finish this up, but I want to know if you're still fond of unittest (and not switched to pytest or something :grin:)

taleinat:
@ionelmc Hi, yes I still prefer unittest because it has the best support for TestCase abstract base classes, which I find myself using quite often.

Apologies for not finishing this myself!

ssbarnea:
close it

PR #53, opened 2016-03-08
force isort to check the correct files and directories

luzfcb:
force isort to check the correct files and directories (this avoid false positive for me)


ionelmc:
Is this really necessary? Tox don't change CWD by default, right?

ionelmc:
@luzfcb hey, is this PR still necessary?

luzfcb:
@ionelmc Not in tox >= 2.3.1

PR #64, opened 2016-05-27
Template fix: missing entry point if "argparse" and config for "nosetests"

tantale:
I change the ``setup.py`` to generate the console script entry point in "argparse" is chosen (not only "plain" or "click").

I add a section for ``nosetests`` in ``setup.cfg`` if "nose" if chosen and remove the ``pytest`` section.

I fix the ``nosetests`` command in ``tox.ini``: "tests" is removed because it is interpreted as being a package instead of a directory.

### DIFF
--- a/{{cookiecutter.repo_name}}/tox.ini
+++ b/{{cookiecutter.repo_name}}/tox.ini
@@ -65,10 +65,10 @@ commands =
     {%- endif %}
 {%- else %}
     {%- if cookiecutter.test_matrix_separate_coverage|lower == 'yes' %}
-    nocov: {posargs:nosetests -v tests}
+    nocov: {posargs:nosetests -v}
     cover: {posargs:nosetests --with-coverage --cover-package={{ cookiecutter.package_name|replace('-', '_') }}}
     {%- else %}
-    {posargs:nosetests --with-coverage --cover-package={{ cookiecutter.package_name|replace('-', '_')}} tests}
+    {posargs:nosetests --with-coverage --cover-package={{ cookiecutter.package_name|replace('-', '_')}}}
     {%- endif %}
 {%- endif %}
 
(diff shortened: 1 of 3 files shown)


{{cookiecutter.repo_name}}/tox.ini
```diff
@@ -65,10 +65,10 @@ commands =
     {%- endif %}
 {%- else %}
     {%- if cookiecutter.test_matrix_separate_coverage|lower == 'yes' %}
-    nocov: {posargs:nosetests -v tests}
```
ionelmc: This needs not be removed: it would collect the code in the `src` dir while normally it's imported from `site-packages`. That would lead to many subtle issues(like doubly executed module, reference issues, collection errors etc).

ionelmc: Hmmm... on second thought, does the `[nosetests]where` in `setup.cfg` always apply?

ionelmc:
Good catch about that missing entrypoint. I wonder if the default tests should be integration tests (the missing entrypoint would had been caught early) ...

{{cookiecutter.repo_name}}/tox.ini
tantale: I fact ``nosetes`` read the ``setup.cfg`` if it is started in the project's directory.
See: http://nose.readthedocs.io/en/latest/usage.html#configuration
Is the current working directory set during ``tox`` task?

ionelmc: CWD isn't changed. It's fine, however, if the location is fixed in the config file then the `cover` envs have no change of collecting doctests. This is why it was specified on command line. Less than ideal but `test_matrix_separate_coverage` is not for everyone either :)

Issue #69, opened 2016-12-14
bootstrap env is missing basepython


rfleschenberg:
The generated `tox.ini` doesn't have `basepython` set for the `bootstrap` environment, so `tox -e bootstrap` doesn't work as documented. An easy workaround is to just run `python ci/bootstrap.py` directly.

Steps to reproduce:

* `mkvirtualenv test1`
* `pip install tox cookiecutter`
* `curl https://raw.githubusercontent.com/rfleschenberg/rf-django-misc/master/.cookiecutterrc > testconfig`
* `cookiecutter --config-file=testconfig gh:ionelmc/cookiecutter-pylibrary`
* Accept defaults for all prompts
* `cd rf-django-misc`
* `tox -e bootstrap`

```
GLOB sdist-make: /tmp/rf-django-misc/setup.py
bootstrap recreate: /tmp/rf-django-misc/.tox/bootstrap
ERROR: InterpreterNotFound: 
_______________________________________________________ summary _______________________________________________________
ERROR:   bootstrap: InterpreterNotFound: 
```

rfleschenberg:
@ionelmc I'm closing this issue, I think it's fixed with #70?

PR #70, opened 2016-12-14
Set basepython for the bootstrap env

rfleschenberg:
Refs #69

### DIFF
--- a/{{cookiecutter.repo_name}}/ci/templates/tox.ini
+++ b/{{cookiecutter.repo_name}}/ci/templates/tox.ini
@@ -11,7 +11,7 @@ envlist =
 [testenv]
 basepython =
     {docs,spell}: python2.7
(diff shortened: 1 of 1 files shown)


ionelmc:
Good catch,  thanks!

rfleschenberg:
I think the build failures are unrelated to my change. I have been thinking of ensuring CI build determinism by pinning all the stuff, by the way. Any thoughts on this?

ionelmc:
I'm using pins in various other projects and it turns out it's a lot of churn to upgrade the pins. It's nice to have repeatable builds but I don't mind merging stuff that breaks build - move fast and break things they say 😁 

On the other hand the few deps this template includes are not breaking that bad, most of the pain is around new/changed style checks in flake8 - I'm not sure pinning that is the right thing ... feel free to argue if you have an opinion.

PR #71, opened 2016-12-14
Fix twine examples in README

rfleschenberg:
Twine doesn't take multiple packages as arguments. Also, you will have `docs` in `dist` after following the README. So the example twine commands don't work.

I guess the commands from this PR will not work with the standard Windows shell. Not sure if you want to support that, or if we can assume a UNIX shell.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -382,8 +382,8 @@ And then you can build the ``sdist``, and if possible, the ``bdist_wheel`` too::
 
 To make a release of the project on PyPI, assuming you got some distributions in ``dist/``, the most simple usage is::
 
-    twine register dist/*
-    twine upload --skip-existing dist/*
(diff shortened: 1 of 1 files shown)


README.rst
```diff
@@ -382,8 +382,8 @@ And then you can build the ``sdist``, and if possible, the ``bdist_wheel`` too::
 
 To make a release of the project on PyPI, assuming you got some distributions in ``dist/``, the most simple usage is::
 
-    twine register dist/*
-    twine upload --skip-existing dist/*
```
ionelmc: Isn't `twine upload --skip-existing dist/*.gz dist/*.whl` simpler?

rfleschenberg: You are right. I didn't realize that `twine upload` does take multiple arguments.

What about the `register` command? Do we  have to run that for both the wheel and the tar.gz, or only one of them?

PR #72, opened 2017-01-10
Update years: add 2016-2017.

hiway:
Noticed that the range 2016-2017 needs to be inserted in the list.


ionelmc:
Good catch, thanks.

PR #74, opened 2017-01-31
Fix redirect warning for tox.readthedocs.org -> tox.readthedocs.io 

hiway:
Running `$ tox -e docs` warns about a permanent redirect:

```
writing output... [ 33%] contributing
...
(line   52) redirect  http://tox.readthedocs.org/en/latest/install.html\
 - permanently to http://tox.readthedocs.io/en/latest/install.html
...
```

This change fixes it:

```
(line   52) ok        http://tox.readthedocs.io/en/latest/install.html
```


ionelmc:
Good catch, thanks!

Issue #77, opened 2017-03-20
travis yml formatting issue


frostidaho:
It looks like there is some issue in the travis.yml template

https://gist.github.com/frostidaho/828e018131a6f8e72a24fa2182e99d95

ionelmc:
Oooooops, I had this issue in the past. Hmmmm ... what jinja version do you have?

frostidaho:
2.9.5

ionelmc:
Should be fine now, everything green at https://travis-ci.org/ionelmc/python-nameless/branches

frostidaho:
thanks :)

PR #78, opened 2017-03-20
Patch for issue #77

frostidaho:
So this fixes the issue I was having. Though I'm not sure if it's entirely correct
#77


ionelmc:
Commited this partially. Thanks!

PR #85, opened 2017-06-30
Add support for setup.py tests_require

techdragon:
- Add the tests_require field to setup.py

### DIFF
--- a/{{cookiecutter.repo_name}}/setup.py
+++ b/{{cookiecutter.repo_name}}/setup.py
@@ -129,6 +129,13 @@ setup(
         # eg: 'aspectlib==1.1.1', 'six>=1.7',
 {%- endif %}
     ],
+    tests_require=[
+{%- if cookiecutter.test_runner|lower == 'pytest' %}
+        'pytest',
+{%- else %}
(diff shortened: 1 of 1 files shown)


ionelmc:
Hmmm, where is this useful? People who run `setup.py test`? Wouldn't that need something like https://pypi.python.org/pypi/pytest-runner?

techdragon:
Ideally, I'd like to see Tox use the setup.py test_requires to add any additional test dependencies automatically when building the tox environments, instead of having to add them to the tox environment configuration. This would reduce some duplication of configuration information in the overall project structure. 

IDE's like PyCharm can use this particular setting in setup.py to ensure that a working environment or virtualenv, has the packages necessary to run the tests and help correctly configure a debugging environment. While they are not able to do the same to packages added into the Tox configuration options.

techdragon:
@ionelmc #122 added `setup_requires` not `tests_require` so that PR has not actually closed this one.

ionelmc:
@techdragon that PR aliases `setup.py test` to a different command provided by pytest-runner. Isn't that enough?

PR #87, opened 2017-07-12
Fix license not being set correctly in setup script

insertjokehere:
Regardless of what license you pick in the cookiecutter dialog, the resulting setup.py script specifies the license as "BSD"


ionelmc:
Thanks!

PR #90, opened 2018-02-13
Update .gitignore

pmav99:
Update .gitignore with the latest definition from https://www.gitignore.io/api/python

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,37 +1,108 @@
+
+# Created by https://www.gitignore.io/api/python
+
+### Python ###
+# Byte-compiled / optimized / DLL files
+__pycache__/
 *.py[cod]
+*$py.class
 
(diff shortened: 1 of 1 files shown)


ionelmc:
I like the idea but some of that stuff is excessive (eg: ignoring all directories named target, var, wheels, env, vent, instance).

I would rather have this grow organically (someone has a slightly different workflow and submits a PR with a couple lines of the extra stuff) than just cargo-cult other people's boilerplate.

pmav99:
Honestly speaking, I don't think that it is really possible to reach a consensus on .gitignores :)

I guess that whether somebody prefers a "fat" or a "slim" .gitignore is probably a matter of taste. Both approaches have their merit, and you will probably need to edit both of them anyway.

Closing the issue.

PR #91, opened 2018-03-16
Added support for unittest

cmarquardt:
Based on Taleinat's pull request for the original project, plus added working unit tests for unittest.


cmarquardt:
Oops - sorry - that wasn't supposed to arrive here, but in my own repo. Closing again...

PR #107, opened 2018-12-10
Update PyPI URLs (PyPI migrated to pypi.org)

janLuke:
https://pypi.python.org/pypi/<pkg-name> redirects permanently to https://pypi.org/project/<pkg-name>


ionelmc:
Thanks!

PR #109, opened 2019-01-06
Fix deprecation warning of `setup.py check`

### LINKED ISSUE #108: Deprecation warning when using `setup.py check` through tox
Hi, first of all thanks for the great project!
When running tox from my bootstrapped project, I have the following warning that gets printed:
`warning: Check: This command has been deprecated. Use `twine check` instead: https://packaging.python.org/guides/making-a-pypi-friendly-readme#validating-restructuredtext-markup`

Its not a big deal but I thought it was worth to be fixed, I'll push a PR shortly

bagerard:
Inspired by how it got fixed in https://github.com/pyca/cryptography/issues/4523
Fixes #108


janLuke:
This always gives an error after you've produced the doc in dist/docs, right?

`Checking distribution dist/docs: InvalidDistribution: Unknown distribution format: 'docs'`

ionelmc:
Yeah dist/docs is a problem. I wish there was another way of doing this. Twine, an upload tool, *is* the wrong place to do this check.

ionelmc:
Either way, even if twine could be the right tool the check env is not meant to build dists, and twine check would only test one of the dists - which is inadequate.

To really solve this pylibrary needs a way to automate publishing (and then piggyback twine check on top of that).

ionelmc:
Btw it was deprecated in this ninja change: https://github.com/pypa/readme_renderer/pull/122 - devoid of any discussion or explanation.

ionelmc:
This is still on my todo list. Just need to figure out a clean way do do it cross-platform (like using some tmp dir) without mucking about in the dist folder which might have other dists there.

Querela:
> This always gives an error after you've produced the doc in dist/docs, right?
> 
> `Checking distribution dist/docs: InvalidDistribution: Unknown distribution format: 'docs'`

You could always specify a file extension pattern, and then it will work ...
E.g.: `twine check dist/*.{whl,gz}`
I found that `twine check` is slightly more strict than `python setup.py check [...]`.

PR #117, opened 2019-07-24
Remove redundant settings in tox.ini

dHannasch:
This removes a [testenv:bootstrap] section in tox.ini that appears (?) to be completely redundant, duplicating the [testenv:bootstrap] above it so that those settings are in two places (so that changing the settings above has no effect, because they are overridden by the settings below).

The only settings left inside the if-block are the settings that actually depend on if cookiecutter.test_matrix_configurator == "yes".

This uses #118 to pass the Travis CI build.

### DIFF
--- a/{{cookiecutter.repo_name}}/tox.ini
+++ b/{{cookiecutter.repo_name}}/tox.ini
@@ -1,12 +1,4 @@
-{%- if cookiecutter.test_matrix_configurator == "yes" -%}
-; this is just a bootstrap tox configuration, run tox or ./ci/bootstrap.py to have the full tox.ini
-
-[tox]
-envlist = bootstrap
-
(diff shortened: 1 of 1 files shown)


ionelmc:
Hmmm, I might have rebased the other pr and caused conflict?

ionelmc:
Hmmm it still has the conflict? Will you rebase?

dHannasch:
> Hmmm it still has the conflict? Will you rebase?

I just rebased this onto ionelmc/master. It looks like that's fixing things.

dHannasch:
Hmm. The 82nd build says simply "An error occurred while generating the build script." with no further information. I don't know what to make of that.

ionelmc:
Travis is just flaky.

PR #116, opened 2019-05-29
Fix incorrect date

jaydenmilne:
Not all of us have time machines :)

PR #118, opened 2019-07-24
Attempt to fix ERROR: unknown environment 'cythonize'

### LINKED ISSUE #119: Builds 14, 15, 32 and 33 fail with ERROR: unknown environment 'cythonize'
In the Travis CI, the 14th, 15th, 32nd and 33rd builds fail on tox -e cythonize with the error message ERROR: unknown environment 'cythonize'. See e.g. https://travis-ci.org/ionelmc/cookiecutter-pylibrary/builds/563150269?utm_source=github_status&utm_medium=notification.

It looks like those four are exactly the builds with test_matrix_configurator: 'yes' that use the environment [testenv:cythonize].

Looking at https://github.com/ionelmc/cookiecutter-pylibrary/blob/master/%7B%7Bcookiecutter.repo_name%7D%7D/tox.ini, it appears that almost the entire file (after line 18) is void if cookiecutter.test_matrix_configurator == "yes". So if cookiecutter.test_matrix_configurator == "yes", then [testenv:cythonize] is never defined.

Moving the definition of [testenv:cythonize] outside the {%- if cookiecutter.test_matrix_configurator == "yes" -%} block appears to fix things; see #118.

(This appears to be unrelated to #27.)

dHannasch:
Resolves #119.

This only makes the Travis CI pass; the AppVeyor build still fails the same way as in e.g. #116. (Which might be related to https://github.com/tox-dev/tox/issues/360, but I don't know.)

### DIFF
--- a/{{cookiecutter.repo_name}}/tox.ini
+++ b/{{cookiecutter.repo_name}}/tox.ini
@@ -114,6 +114,8 @@ commands =
     isort --verbose --check-only --diff --recursive src tests setup.py
 {%- if cookiecutter.c_extension_support == 'cython' %}
 
+; This should define the test environment 'cythonize',
+; for when we later subprocess.check_call(['tox', '-e', 'cythonize']).
 [testenv:cythonize]
 basepython = {env:TOXPYTHON:python}
 deps =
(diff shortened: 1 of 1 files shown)


ionelmc:
Good catch, thanks!

PR #120, opened 2019-07-24
bootstrap tox

### LINKED ISSUE #121: bootstrap fails if tox is not in PATH
`cookiecutter gh:ionelmc/cookiecutter-pylibrary` crashes if tox is not available on the command line.
Tracing the exception back to https://github.com/ionelmc/cookiecutter-pylibrary/blob/master/%7B%7Bcookiecutter.repo_name%7D%7D/ci/bootstrap.py#L67, it appears this was previously known.
It seems like we could bootstrap tox by adding it to the dependencies, as in #120.

dHannasch:
Resolves #121.

Since tox is required by the bootstrap.py script (to run tox --listenvs), this adds tox to the dependencies and then invokes tox the same way it's invoked in https://github.com/ionelmc/cookiecutter-pylibrary/blob/master/hooks/post_gen_project.py.

This relies on #117.

### DIFF
--- a/{{cookiecutter.repo_name}}/tox.ini
+++ b/{{cookiecutter.repo_name}}/tox.ini
@@ -2,6 +2,7 @@
 deps =
     jinja2
     matrix
+    tox
 skip_install = true
 commands =
     python ci/bootstrap.py
--- a/{{cookiecutter.repo_name}}/ci/bootstrap.py
+++ b/{{cookiecutter.repo_name}}/ci/bootstrap.py
@@ -64,8 +64,12 @@ if __name__ == "__main__":
 {% else %}
     tox_environments = [
         line.strip()
-        # WARNING: 'tox' must be installed globally or in the project's virtualenv
-        for line in subprocess.check_output(['tox', '--listenvs'], universal_newlines=True).splitlines()
+        # 'tox' need not be installed globally, but must be importable
+        # by the Python that is running this script.
+        # This uses sys.executable the same way that the call in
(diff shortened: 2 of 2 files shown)


{{cookiecutter.repo_name}}/tox.ini
```diff
@@ -1,20 +1,24 @@
...
+    tox
 skip_install = true
 commands =
     python ci/bootstrap.py
+{% if cookiecutter.test_matrix_configurator == "yes" %}
+basepython =
+    python
 passenv =
     *
+{% endif %}
```
ionelmc: I think the endif should be before the passenv.

ionelmc:
Anyway, if you want you can make a single pr with all the changes you want - I think that will make things simpler (I'm not fussy about small PRs)

{{cookiecutter.repo_name}}/tox.ini
dHannasch: Fair enough. I didn't have any particular reason for passenv to be specific to if cookiecutter.test_matrix_configurator == "yes", I just put it there because previously it *was* specific to if cookiecutter.test_matrix_configurator == "yes" and I figured maybe there was a reason.

dHannasch:
> Anyway, if you want you can make a single pr with all the changes you want - I think that will make things simpler (I'm not fussy about small PRs)

Fair enough. I just rebased this onto #117, with one commit for reorganizing tox.ini and one commit for bootstrapping tox.

ionelmc:
Took the liberty of rebasing it without the other pr's changes.

PR #122, opened 2019-07-24
Use test runner for setup py test

dHannasch:
This is sort of a feature request: This changes setup.py and setup.cfg so that in a newly-created package, by default, python setup.py test will run whatever test runner was selected.

This required collapsing multiple settings of `setup_requires` in setup.py so that we could include the necessary pytest-runner and still include cffi and cython in `setup_requires` when needed. (I can split of that collapsing as a separate branch if that would be clearer.)

The best way I could think of to allow the jinja2 if-blocks to either add or not add to `setup_requires` was to use list concatenation.

### DIFF
--- a/{{cookiecutter.repo_name}}/setup.py
+++ b/{{cookiecutter.repo_name}}/setup.py
@@ -167,18 +167,17 @@ setup(
         #   ':python_version=="2.6"': ['argparse'],
     },
     tests_require=['{{cookiecutter.test_runner}}'],
-    setup_requires=(
-        []
+    setup_requires=list(filter(None, [
 {%- if cookiecutter.test_runner == 'pytest' %}
-        + ['pytest-runner']
+        'pytest-runner',
 {% endif %}
 {%- if cookiecutter.c_extension_support == 'cython' %}
-        + (['cython'] if Cython else [])
+        ['cython'] if Cython else None,
 {% endif %}
 {%- if cookiecutter.c_extension_support == 'cffi' %}
-        + (['cffi>=1.0.0'] if any(i.startswith('build') or i.startswith('bdist') for i in sys.argv) else [])
+        ['cffi>=1.0.0'] if any(i.startswith('build') or i.startswith('bdist') for i in sys.argv) else None,
 {% endif %}
-    ),
+    ])),
 {%- if cookiecutter.command_line_interface != 'no' %}
     entry_points={
         'console_scripts': [
--- a/{{cookiecutter.repo_name}}/setup.py
+++ b/{{cookiecutter.repo_name}}/setup.py
@@ -167,15 +167,25 @@ setup(
         #   ':python_version=="2.6"': ['argparse'],
     },
 {%- if cookiecutter.test_runner == 'pytest' -%}
-{% set cookiecutter.setup_requires_from_test_runner %}'pytest-runner', {% endset %}
+{% set setup_requires_from_test_runner %}
+        'pytest-runner',{% endset %}
 {%- else -%}
-{% set cookiecutter.setup_requires_from_test_runner %}{% endset %}
+{% set setup_requires_from_test_runner %}{% endset %}
 {%- endif -%}
 {%- if cookiecutter.c_extension_support == 'cython' %}
-    setup_requires = [{{cookiecutter.setup_requires_from_test_runner}}'cython'] if Cython else [{{cookiecutter.setup_requires_from_test_runner}}]
+    setup_requires = [{{setup_requires_from_test_runner}}
+        'cython',
+    ] if Cython else [{{setup_requires_from_test_runner}}
+    ]
 {%- elif cookiecutter.c_extension_support == 'cffi' %}
-    setup_requires = [{{cookiecutter.setup_requires_from_test_runner}}'cffi>=1.0.0'] if any(i.startswith('build') or i.startswith('bdist') for i in sys.argv) else [{{cookiecutter.setup_requires_from_test_runner}}],
-{% endif %}
+    setup_requires = [{{setup_requires_from_test_runner}}
+        'cffi>=1.0.0',
+    ] if any(i.startswith('build') or i.startswith('bdist') for i in sys.argv) else [{{setup_requires_from_test_runner}}
+    ],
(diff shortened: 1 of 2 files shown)


{{cookiecutter.repo_name}}/setup.py
```diff
@@ -166,11 +166,19 @@ def _unavailable(self, e):
         #   'rst': ['docutils>=0.11'],
         #   ':python_version=="2.6"': ['argparse'],
     },
+    tests_require=['{{cookiecutter.test_runner}}'],
+    setup_requires=(
+        []
```
ionelmc: I think this chaining will make the template repeat less ifs but the output would be ugly. Can we have something like this?
```
{%- if cookiecutter.test_runner == 'pytest' %}
 (['pytest-runner', 'cython'] if Cython else ['pytest-runner'])
{% else %}
 (['cython'] if Cython else [])
{% endif %}
```

Also, test_requires should also have test deps. Perhaps place a comment there to help users figure out the right place for test deps? (there would be two places, tox.ini and test_requires unfortunately, user should choose one style of doing it)

ionelmc:
Just a note for me (this will close #85).

{{cookiecutter.repo_name}}/setup.py
dHannasch: I certainly can't deny that the chained concatenation of lists looks clunky in the produced `setup.py`, but it seems like mashing the if-blocks together would make the logic very hard to follow.

> ```
> {%- if cookiecutter.test_runner == 'pytest' %}
>  (['pytest-runner', 'cython'] if Cython else ['pytest-runner'])
> {% else %}
>  (['cython'] if Cython else [])
> {% endif %}
> ```

The bit you posted looks fine on its own, but remember that it'll have to be wrapped in an if-block for Cython followed by another, not-identical-but-must-be-kept-in-sync-in-non-obvious-ways if-block for cffi...

If I'm hearing you correctly, ultimately we want the generated `setup_requires` to look like a normal list written by hand, right? That's what makes it easiest to edit and add to.

And of course with jinja2, we can just include and exclude lines easily. The catch comes when some entries should be included or excluded from the list not only based on cookiecutter settings known at cutting-time, but *also* based on things known only at runtime, like sys.argv.

So we want a list that looks basically like a normal list, *but* the cython and cffi entries are included or excluded based on criteria at runtime.

Hmm. What if the cython and cffi entries were always included in the list (if cookiecutter.c_extension_support is set) but might be None? Same idea as making them sometimes the empty list, but without the concatenation.
Unfortunately setuptools does not ignore None in `setup_requires`, so we'd have to slap a filter on the front of the list. That's a little ugly at the start, but easy to edit and add to in the generated setup.py --- I would think most people would find it intuitive that `else None` would mean that requirement being left out, even before squinting at the `list(filter(` to understand exactly what it's doing.

ionelmc: Yes, I want output to look "compact handwritten" like. I think there's no way around repeating lots of stuff in template, sorry :/

Fusing the cython and cffi support would simplify the template but at the end people will ask this: "why would I use a template that generates such verbose and ugly output when there are dozen cleaner alternatives?"

dHannasch: Hmm, okay. So you want a skeleton like this, correct?

```
{%- if cookiecutter.c_extension_support == 'cython' %}
    setup_requires = ...
{%- elif cookiecutter.c_extension_support == 'cffi' %}
    setup_requires = ...
{%- else %}
    setup_requires = ...
{%- endif -%}
```

And then handle the rest in there somehow, possibly with nested if-blocks in each.

```
{%- if cookiecutter.c_extension_support == 'cython' %}
    setup_requires =
{%- if cookiecutter.test_runner == 'pytest' %}
 (['pytest-runner', 'cython'] if Cython else ['pytest-runner'])
{% else %}
 (['cython'] if Cython else [])
{% endif %}
{%- elif cookiecutter.c_extension_support == 'cffi' %}
    setup_requires = ...
{%- else %}
    setup_requires = ...
{%- endif -%}
```

What about setting a variable right before, so that the if-block isn't copied three times?

```
{%- if cookiecutter.test_runner == 'pytest' -%}
{% set setup_requires_from_test_runner %}
        'pytest-runner',{% endset %}
{%- else -%}
{% set setup_requires_from_test_runner %}{% endset %}
{%- endif -%}
```

(Of course this could also be done with a macro.)

Incidentally, I'm guessing we want one requirement per line to be consistent with the other list arguments in https://github.com/ionelmc/python-nameless/blob/master/setup.py?

```
    setup_requires = [
        'pytest-runner',
        'cython',
    ],
```

ionelmc: No preference for template style (anything you like is fine). I threw in the towel in making pretty jinja templates some time ago 🤕 

Make the output the best you can, all I can say.

dHannasch: Does the version I just pushed look good, then? The output looks like

```
    setup_requires = [
        'pytest-runner',
        'cython',
    ] if Cython else [
        'pytest-runner',
    ],
```

ionelmc: It's perfect. I'm gonna try build some projects using your branch and then I'll merge.

dHannasch: > I'm gonna try build some projects using your branch and then I'll merge.

Yeah, there's always that awkward step of "well, the tests pass, but what does the generated python-nameless actually look like?" It's too bad there's no way to view the projects that get cookiecut by Travis for the tests --- it generates a bunch of projects, but there's no way to simply look at a couple of them the way you can look at https://github.com/ionelmc/python-nameless. Sadly there seems to be no way to get previews like that: https://github.com/travis-ci/dpl/issues/991

ionelmc:
I made some changes to support offline use (a test_runner_setup_requires option).

ionelmc:
@dHannasch fyi https://github.com/pypa/setuptools/issues/1684 :\

ionelmc:
Yes fair point. I guess just be subscribed to that issue for now.

PR #123, opened 2019-07-25
Run installable tests

dHannasch:
Minor suggestion for pytest default settings --- going for least astonishment, someone who puts a `test_*` file in `src/` (that is, a test to be installed with the package) probably will not be surprised or confused if `pytest` finds and runs those tests.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -225,6 +225,12 @@ You will be asked for these fields:
       - Whether to use the test_runner for python setup.py test.
         Note that this will also add to ``setup_requires`` if a test-runner is needed.
 
+    * - ``run_tests_that_are_installed_with_package``
+      - .. code:: python
+
+            "pytest"
+      - Run tests in the source directory (that is, tests installed with the package).
+
     * - ``linter``
       - .. code:: python
 
--- a/README.rst
+++ b/README.rst
@@ -225,10 +225,10 @@ You will be asked for these fields:
       - Whether to use the test_runner for python setup.py test.
         Note that this will also add to ``setup_requires`` if a test-runner is needed.
 
-    * - ``tests_installed_with_package_also_run``
+    * - ``allow_tests_inside_package``
       - .. code:: python
 
-            "pytest"
+            "no"
       - Run tests in the source directory (that is, tests installed with the package).
 
     * - ``linter``
--- a/README.rst
+++ b/README.rst
@@ -225,7 +225,7 @@ You will be asked for these fields:
       - Whether to use the test_runner for python setup.py test.
         Note that this will also add to ``setup_requires`` if a test-runner is needed.
 
-    * - ``run_tests_that_are_installed_with_package``
+    * - ``tests_installed_with_package_also_run``
       - .. code:: python
 
             "pytest"
--- a/cookiecutter.json
+++ b/cookiecutter.json
@@ -42,6 +42,7 @@
     "test_matrix_separate_coverage": ["no", "yes"],
     "test_runner": ["pytest", "nose"],
     "test_runner_used_for_setuppy_test": ["no", "yes"],
+    "run_tests_that_are_installed_with_package": ["no", "yes"],
     "linter": ["flake8", "pylama"],
     "command_line_interface": ["plain", "argparse", "click", "no"],
     "command_line_interface_bin_name": "{{ cookiecutter.distribution_name }}",
--- a/cookiecutter.json
+++ b/cookiecutter.json
@@ -42,7 +42,7 @@
     "test_matrix_separate_coverage": ["no", "yes"],
     "test_runner": ["pytest", "nose"],
     "test_runner_used_for_setuppy_test": ["no", "yes"],
-    "tests_installed_with_package_also_run": ["no", "yes"],
+    "allow_tests_inside_package": ["no", "yes"],
     "linter": ["flake8", "pylama"],
     "command_line_interface": ["plain", "argparse", "click", "no"],
     "command_line_interface_bin_name": "{{ cookiecutter.distribution_name }}",
--- a/cookiecutter.json
+++ b/cookiecutter.json
@@ -42,7 +42,7 @@
     "test_matrix_separate_coverage": ["no", "yes"],
     "test_runner": ["pytest", "nose"],
     "test_runner_used_for_setuppy_test": ["no", "yes"],
-    "run_tests_that_are_installed_with_package": ["no", "yes"],
+    "tests_installed_with_package_also_run": ["no", "yes"],
     "linter": ["flake8", "pylama"],
     "command_line_interface": ["plain", "argparse", "click", "no"],
     "command_line_interface_bin_name": "{{ cookiecutter.distribution_name }}",
--- a/hooks/post_gen_project.py
+++ b/hooks/post_gen_project.py
@@ -41,6 +41,9 @@ if __name__ == "__main__":
 {%- if cookiecutter.test_matrix_configurator == 'no' %}
     os.unlink(join('ci', 'templates', 'tox.ini'))
 {% endif %}
+{%- if cookiecutter.tests_installed_with_package_also_run == 'no' %}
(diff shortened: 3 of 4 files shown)


ionelmc:
Unfortunately this will break everything in practice because there would be two unpredictable sources for code - src and the stuff in site-packages.

For people that need to collect test in src there is the test_matrix_separate_coverage option.

ionelmc:
Or without test_matrix_separate_coverage you'd basically need to have `testpaths = tests path/to/site-packages/mypackage` which is not adequate from the get-go.

dHannasch:
> there would be two unpredictable sources for code - src and the stuff in site-packages.

If I'm understanding you correctly, the problem is that after making a point of isolating code in `src/` so that we're forced to test the installed version of the code (https://blog.ionelmc.ro/2014/05/25/python-packaging/#the-structure), we're not testing the installed version of the *tests*, correct?

That's a good point. What if we used --pyargs to run the installed version of the tests?

```
addopts =
    --pyargs
testpaths =
    {{cookiecutter.package_name}}
```

I realize it looks strange, but it appears this really is the intended interaction between pyargs and testpaths, and it got fixed when 4.0.0 briefly broke it: https://github.com/pytest-dev/pytest/issues/4405

Someone suggested allowing package names to be specified in testpaths itself (https://github.com/pytest-dev/pytest/issues/3405#issuecomment-382377903), but that never got off the ground since technically that behavior is already possible in pytest (albeit strange-looking).

ionelmc:
Hmmm very interesting (and what a horrible choice for option name). We could have a "tests_inside_package" or "package_contains_test" if you fancy the idea of waddling around in the templates again :)

ionelmc:
I mean, in the idea that adding `--pyargs` will mess up usage of "tests outside" and it's not a good default (I guess there would be some surprise factor when playing with extra collection options, and tests that do crazy stuff at import time blowing up when tools try to inspect installed modules).

dHannasch:
> We could have a "tests_inside_package" or "package_contains_test"
> adding `--pyargs` will mess up usage of "tests outside" and it's not a good default (I guess there would be some surprise factor when playing with extra collection options, and tests that do crazy stuff at import time blowing up when tools try to inspect installed modules).

I have no objection to making a run_installable_tests option per se, but I'm a bit confused by your comment.

I'm not sure what you mean by mess up usage of tests outside? This still runs the non-installed tests as normal: https://github.com/ionelmc/cookiecutter-pylibrary/pull/123/files#diff-9187e1a09582306a1c23d6b37065dffdR52

I put a / at the end so even if somebody has installed a package literally named tests (https://pypi.org/project/tests/), pytest will still run the tests in the ./tests/ directory because it can't parse "tests/" as a package name.

(Though now that I think about it, I'll have to check whether that works on Windows...sometimes Python tools will accept / paths on Windows https://github.com/pallets/jinja/issues/411#issuecomment-230678365 , but I don't know about pytset testpaths. Might need something to put a \ there on Windows. Looks like there are ways, and probably / is fine in any event: https://github.com/pallets/jinja/issues/367)

I don't see any way that this could interfere with running the "outside" non-installed tests. Is there a possibility I'm overlooking?

We're definitely going to have to import the installed {{cookiecutter.package_name}} to run any tests at all, including the ones in tests/ --- indeed we went out of our way to make sure we were forced to import the installed package when running the tests, because we specifically wanted that (https://blog.ionelmc.ro/2014/05/25/python-packaging/#the-structure).

I would be very surprised if anyone was surprised by the installed tests being run, considering that in order for any installed tests to get run, the user would first have to *write* those tests. Someone who sits down to write a test presumably intends to run it, I would think. Even if --- especially if --- they're making that test available for their own users to run with `pytest --pyargs {{cookiecutter.package_name}}`.

Does anything unusual happen when tools try to inspect installed modules? I mean...I'm not an expert on Python packaging, but I thought an installed package was literally just a folder.

My biggest concern is people playing with extra collection options you mentioned --- I'm not very familiar with all those. I wouldn't think that any of the test-collection options pytest has would do anything crazy like "escape" the package folder and go running around elsewhere in site-packages, but it sounds like you'd know more about that than I would. Are there any pytest collection options you're thinking of that would go and do crazy things if we were running tests out of an installed package?

ionelmc:
I guess this is relevant: https://github.com/pytest-dev/pytest/issues?utf8=%E2%9C%93&q=is%3Aissue+is%3Aopen+pyargs and I suspect there are more unknown ones because few people use the obscure `--pyargs`.

It's basically gut feeling + the fact that other "late and not really wanted features" of pytest like unittest support have various problems and are not orthogonal to other pytest features (won't work well or integrate well with other features or CLI options).

Regardless of technical implications, can't just switch test layout overnight. Maybe later we can change the default - if it work well. 

You agree on having an option for this?

ionelmc:
Just one small nit, in the vein of the discussion on the other PR - the option prefixing (I would like to have the option start with "test") :) 
 
Some suggestions:
- tests_installed_with_package
- tests_inside_package
- tests_in_package
- tests_installable
- test_installed_package
- test_without_sources
- test_without_repo

dHannasch:
> I guess this is relevant: https://github.com/pytest-dev/pytest/issues?utf8=%E2%9C%93&q=is%3Aissue+is%3Aopen+pyargs and I suspect there are more unknown ones because few people use the obscure `--pyargs`.

Yikes. That's a lot. Since pyargs has been around so long, I always sort of assumed it was just poorly documented, not buggy.

What do you think of the name `tests_installed_with_package_also_run`?

Hmm. `The command "openssl aes-256-cbc -K $encrypted_a70d5afca909_key -iv $encrypted_a70d5afca909_iv -in publish-key.enc -out ~/.ssh/publish-key -d" exited with 1.` Probably related to https://github.com/ionelmc/cookiecutter-pylibrary/commit/7f5dd2cd91f2e54f7fb0ffb46a5a21e5abd86d6c and the bug you were trying to fix with https://github.com/ionelmc/cookiecutter-pylibrary/commit/5de08d335844f048c00039acfa78bbbedfc39c3e . Unfortunately it's not at all clear *why* openssl would crash.

ionelmc:
Not sure what to say about the name. I guess I need to find some time to really try `--pyargs` and see how it behaves in practice, and also overhaul the CI a bit, sigh ...

dHannasch:
For what it's worth, the autodeploy looks incredibly cool (to the extent I understand it, which isn't much). No idea why it isn't working.

`pytest --pyargs {{name}}` is a lifesaver if you do a lot of subprocess shenanigans or low-level hacking that depends on the host machine in unpredictable ways. It'll always find the tests no matter what variety of Python the user has installed or what kind of virtual environment they're using, and the full work-up is usually enough for a good swift guess as to what's different about the environment that's throwing a wrench in matters. (Of course, coming up with a patch to allow the package to work in that environment isn't any easier, but at least you don't have to fiddle around wondering what the problem is.)
If you _don't_ do any such shenanigans, then you can easily go your entire life without wanting or needing pyargs, which is probably why it gets less care.

ionelmc:
@dHannasch so I've played a bit with pyargs, and I think to avoid confusion this option should be conflated with the move of the tests inside the package (eg, have `os.rename('tests', 'src/{pakcage_name}/tests'); create 'src/{pakcage_name}/tests/__init__.py` inside the post gen hook).

You still wanna work on this?

ionelmc:
Mainly the confusion around more lenient test collection. Users won't get any errors if argument is not a path but some package name.

The other reason is that my aim with this template is to produce "hand-written"-like output. No one would use `--pyargs` if they don't have tests inside package, thus this template should produce that sort of output.

And it's not `src/tests/something` but `src/package_name/tests/something`. 

To put this differently, this template should not produce something made halfway (I think making users do more changes to make tests inside package work, like adding collection arguments in various places) but produce something that works end to end - namely a working project with tests inside package - as I understand that's the underlying goal here.

If someone needs huge test files then they should not use this option. This should be made note of somewhere in option description.

ionelmc:
Oook ... fine. About the option name, you like any of these alternatives?

- collect_tests_from_package
- allow_tests_inside_package

PR #125, opened 2019-07-26
Document boolean option to use test_runner for python setup.py test.

dHannasch:
This also changes the name to a hopefully-more-descriptive name.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -218,11 +218,12 @@ You will be asked for these fields:
             "pytest"
       - Test runner to use. Available options: ``pytest`` or ``nose``.
 
-    * - ``use_test_runner_for_setuppy_test``
+    * - ``test_runner_used_for_setuppy_test``
       - .. code:: python
 
             "no"
       - Whether to use the test_runner for python setup.py test.
+        Note that this will also add to ``setup_requires`` if a test-runner is needed.
 
     * - ``linter``
       - .. code:: python
--- a/README.rst
+++ b/README.rst
@@ -218,6 +218,12 @@ You will be asked for these fields:
             "pytest"
       - Test runner to use. Available options: ``pytest`` or ``nose``.
 
+    * - ``use_test_runner_for_setuppy_test``
+      - .. code:: python
+
+            "no"
+      - Whether to use the test_runner for python setup.py test.
+
     * - ``linter``
       - .. code:: python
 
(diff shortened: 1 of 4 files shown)


ionelmc:
So the idea with the funny name was that options regarding same area begin with the same prefix. 

Perhaps `test_runner_use_in_setuppy_test` is better? Tho I picked `test_runner_setup_requires` to make it clear it's will impact the setup_requires metadata. Kinda an advanced packaging topic but at least it makes clear of implications of using it. Really broken about what to use. What are your thoughts?

dHannasch:
> So the idea with the funny name was that options regarding same area begin with the same prefix.

Hmm. I suppose if these might get sorted lexicographically...well, we absolutely need to make sure this option comes right after the test_runner choice, so if we can't enforce ordering the options according to the order in cookiecutter.json, then we need a common prefix.

> I picked `test_runner_setup_requires` to make it clear it will impact the setup_requires metadata.

Hmm. That's a tough question. I hadn't thought about the fact that it wouldn't be obvious that this option would impact setup_requires.

On the one hand, since cookiecutter doesn't allow displaying helptext alongside options the way you can do with argparse, my main thinking is to make it as easy as possible for someone to immediately know which option they want to pick, preferably without even looking at the documentation. "Do you want or need to `use test_runner for python setup.py test`?" It almost sounds like a natural English sentence, you know?

On the other hand, it's also important to be transparent about what the *consequences* are of each choice. And it's certainly not obvious that aliasing setup.py test implies adding something to setup_requires. That definitely qualifies as a point of Python packaging trivia.

On the gripping hand, if someone isn't already deep enough into Python packaging wizardry to know that aliasing setup.py test implies adding something to setup_requires, then they very probably don't *care* about adding something to setup_requires.

So I think maybe the main concern is people picking yes and then opening setup.py and seeing setup_requires and saying "What the heck is this? I didn't ask for this. Do I need this? What happens if I delete this?"

What do you think of adding a comment to setup.py itself? That way, if anyone is looking at setup.py and wondering what setup_requires is and why they have it, there's an immediate answer right there.

```
# If you choose to alias python setup.py test to pytest,
# then you need the pytest-runner in setup_requires to enable that.
```

And maybe a note in the table in README.rst that it'll mess with setup_requires...hmm... (Honestly, it would be ideal to just show a git diff of exactly what each option changes in python-nameless, but the first few ways I can think of to do that all sound really difficult.)

ionelmc:
Ok fine, agree on `test_runner_use_in_setuppy_test` then?

I don't think we need extra comments (setup_requires=['pytest something'] is fairly selfdocumenting imo).

ionelmc:
Very well, thanks!

PR #126, opened 2019-08-26
Wheel publishing

ionelmc:
fyi @dHannasch I renamed one of the options (since setuptools-scm support is in the same area).


ionelmc:
Yeah I need to figure a better test strategy - pushing those branches got me banned on appveyor for abuse lol.

dHannasch:
> pushing those branches got me banned on appveyor

I'm guessing you mean for running the https://github.com/ionelmc/python-nameless/ tests, not the https://github.com/ionelmc/cookiecutter-pylibrary tests? (Since I assume the number of tests for the cookiecutter itself is still the same?)

Hmm. I guess I can see how even though test_nameless.py is trivial, appveyor still has to set up the whole environment for each branch, so they wouldn't like the proliferation of branches. I assume the solution is to just disable appveyor for python-nameless? (Which is too bad, on a level, since it's cool to have verification the .appveyor.yml works, but not testing python-nameless probably isn't a big risk.)

PR #127, opened 2019-08-30
Allow non-readthedocs documentation hosting

dHannasch:
Currently, out of the box, `tox -e docs` crashes if documentation is not hosted on readthedocs, complaining of a broken link.

The adds a way to enter a custom URL if documentation is hosted in a non-standard location such as GitHub Pages or GitLab Pages.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -323,6 +323,12 @@ You will be asked for these fields:
             "no"
       - Add a CodeClimate_ badge in ``README.rst``.
 
+    * - ``sphinx_docs``
+      - .. code:: python
+
+            "yes"
+      - Have Sphinx documentation.
+
     * - ``sphinx_theme``
       - .. code:: python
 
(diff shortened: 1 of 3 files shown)


ionelmc:
Hmmm it looks like I forgot to add the `sphinx_docs` option in the readme. Perhaps we could have the same prefix ... sphinx_docs_hosting? What do you think?

dHannasch:
`sphinx_docs_hosting` sounds good.
I had been thinking about it as independent of Sphinx, but it really isn't --- we aren't going to use anything else for documentation, and the entire option to set hosting makes no sense if the user says no to the docs. So this is really a subordinate option of `sphinx_docs`. If cookiecutter gave us the option of only displaying this option when `sphinx_docs` is yes, that would be ideal. Since that's not possible, using sphinx_docs as a prefix seems like a good way to make clear that this is a subordinate option of `sphinx_docs`.

PR #129, opened 2019-08-30
Allow repo hosting not ending in com

### LINKED ISSUE #128: Allow repo_hosting not ending in .com
I would like to allow repo hosting at hosts not ending in .com --- for example .edu, .org, or .yt.
From grepping through all the usages of repo_hosting, it looks like I could pretty much replace every instance of `{{ cookiecutter.repo_hosting }}.com` with just `{{ cookiecutter.repo_hosting }}` and replace the default options `github` and `gitlab` with `github.com` and `gitlab.com`.

(And replace all the `if cookiecutter.repo_hosting == "github"` checks with `if cookiecutter.repo_hosting == "github.com"`.)

Is there a reason not to do this that I'm not thinking of?

dHannasch:
Resolves #128.

### DIFF
--- a/{{cookiecutter.repo_name}}/README.rst
+++ b/{{cookiecutter.repo_name}}/README.rst
@@ -26,7 +26,7 @@ Overview
 {{ '' }}
     * - package
       - | |version| |wheel| |supported-versions| |supported-implementations|
-{%- if cookiecutter.repo_hosting == "github" %}
+{%- if cookiecutter.repo_hosting == "github.com" %}
         | |commits-since|
 {%- endif %}
 {{ '' }}
@@ -78,7 +78,7 @@ Overview
 .. |version| image:: https://img.shields.io/pypi/v/{{ cookiecutter.distribution_name }}.svg
     :alt: PyPI Package latest release
     :target: https://pypi.org/project/{{ cookiecutter.distribution_name }}
-{% if cookiecutter.repo_hosting == "github" %}
+{% if cookiecutter.repo_hosting == "github.com" %}
 .. |commits-since| image:: https://img.shields.io/{{ cookiecutter.repo_hosting }}/commits-since/{{ cookiecutter.repo_username }}/{{ cookiecutter.repo_name }}/v{{ cookiecutter.version }}.svg
     :alt: Commits since latest release
     :target: https://{{ cookiecutter.repo_hosting }}/{{ cookiecutter.repo_username }}/{{ cookiecutter.repo_name }}/compare/v{{ cookiecutter.version }}...master
--- a/{{cookiecutter.repo_name}}/README.rst
+++ b/{{cookiecutter.repo_name}}/README.rst
@@ -81,7 +81,7 @@ Overview
 {% if cookiecutter.repo_hosting == "github" %}
 .. |commits-since| image:: https://img.shields.io/{{ cookiecutter.repo_hosting }}/commits-since/{{ cookiecutter.repo_username }}/{{ cookiecutter.repo_name }}/v{{ cookiecutter.version }}.svg
     :alt: Commits since latest release
-    :target: https://{{ cookiecutter.repo_hosting }}.com/{{ cookiecutter.repo_username }}/{{ cookiecutter.repo_name }}/compare/v{{ cookiecutter.version }}...master
+    :target: https://{{ cookiecutter.repo_hosting }}/{{ cookiecutter.repo_username }}/{{ cookiecutter.repo_name }}/compare/v{{ cookiecutter.version }}...master
 {% endif %}
 .. |wheel| image:: https://img.shields.io/pypi/wheel/{{ cookiecutter.distribution_name }}.svg
     :alt: PyPI Wheel
--- a/README.rst
+++ b/README.rst
@@ -131,8 +131,8 @@ You will be asked for these fields:
     * - ``repo_hosting``
       - .. code:: python
 
-            "github"
-      - Use ``"no"`` for no hosting (various links will disappear). You can also use ``"gitlab"`` and such but various
+            "github.com"
+      - Use ``"no"`` for no hosting (various links will disappear). You can also use ``"gitlab.com"`` and such but various
         things will be broken (like Travis configuration).
 
     * - ``repo_name``
--- a/ci/envs/matrix-sepcov-pytest-cext-plain-flake8-github.cookiecutterrc
+++ b/ci/envs/matrix-sepcov-pytest-cext-plain-flake8-github.cookiecutterrc
@@ -16,7 +16,7 @@ default_context:
   linter: flake8
   package_name: nameless
   project_name: "Nam\xE9\"'less"
-  repo_hosting: github
+  repo_hosting: github.com
   repo_name: python-nameless
   requiresio: 'yes'
   scrutinizer: 'yes'
--- a/ci/envs/matrix-single-pytest-cext-plain-flake8-github.cookiecutterrc
+++ b/ci/envs/matrix-single-pytest-cext-plain-flake8-github.cookiecutterrc
@@ -16,7 +16,7 @@ default_context:
   linter: flake8
   package_name: nameless
   project_name: "Nam\xE9\"'less"
-  repo_hosting: github
+  repo_hosting: github.com
   repo_name: python-nameless
   requiresio: 'yes'
   scrutinizer: 'yes'
--- a/ci/envs/plain-sepcov-nose-pure-plain-flake8-github.cookiecutterrc
+++ b/ci/envs/plain-sepcov-nose-pure-plain-flake8-github.cookiecutterrc
@@ -16,7 +16,7 @@ default_context:
   linter: flake8
   package_name: nameless
   project_name: "Nam\xE9\"'less"
-  repo_hosting: github
+  repo_hosting: github.com
   repo_name: python-nameless
   requiresio: 'yes'
   scrutinizer: 'yes'
--- a/ci/envs/plain-sepcov-pytest-cext-optional-plain-flake8-github.cookiecutterrc
+++ b/ci/envs/plain-sepcov-pytest-cext-optional-plain-flake8-github.cookiecutterrc
@@ -16,7 +16,7 @@ default_context:
   linter: flake8
   package_name: nameless
   project_name: "Nam\xE9\"'less"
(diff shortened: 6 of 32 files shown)


{{cookiecutter.repo_name}}/README.rst
```diff
@@ -1,7 +1,7 @@
 ========
 Overview
 ========
-{% if cookiecutter.repo_hosting != "no" %}
+{% if cookiecutter.repo_hosting == "github.com" %}
```
ionelmc: So I think this made me make those confusing comments. For the .edu hosting you don't want any badging? Note that some badges will work with gitlab.com, maybe include that as well?

Either way, the pypi badges (wheel, version, supported-versions, supported-implementation) are valid in any context (except when project is "private" and doesn't go on pypi). 

So about private projects, for example I got a cffi binding for some crappy proprietary c stuff, I want to make sure it can never go on pypi by mistake. The `setup.py` has this:
```jinja
{%- if cookiecutter.repo_hosting == "no" %}
        'Private :: Do Not Upload',
{%- endif %}
```

A way to let you use custom domain but still give choice input to most users is to have something like this in cookiecutter.json:
```json
{
"repo_hosting": ["github.com", "gitlab.com", "*custom domain*", "*private*"],
"repo_hosting_custom": "{{ cookiecutteer.repo_hosting if cookiecutteer.repo_hosting != '*custom domain*' else '' }}",
}
```

And then we have something like this in `setup.py`:

```jinja
{%- if cookiecutter.repo_hosting == "*private*" %}
        'Private :: Do Not Upload',
{%- endif %}
```

What do you think?

ionelmc: Another thing that I noticed, some badge urls hardcode "github", eg "https://landscape.io/github/blabla" - maybe the options shouldn't have the ".com" part? But then we need to have some ugly thing for custom domain .... or we add the .tld in "repo_hosting_custom" default value?

ionelmc: Also, "repo_hosting_domain" might be a better name instead of "custom".

SterlingPeet: I like ``repo_hosting_domain`` over ``custom``.  I don't quite have the strongest handle on the badges situation, but I do like getting them in my repo template.  I would say that I find it less problematic to remove non-functioning badges in the ``github.gatech.edu`` repo than to not get them at all.

SterlingPeet: munging and removing badges by hand seems much less challenging than replacing all the repo urls across the project

dHannasch: > For the .edu hosting you don't want any badging?

As far as badging, https://github.com/ionelmc/cookiecutter-pylibrary/pull/129/commits/db74b277d1e6f4d94f6cf2aadd87e8c664d1a266 ports the documentation badge to GitLab Pages, and https://github.com/ionelmc/cookiecutter-pylibrary/pull/135 adds a separate option to disable the PyPI links. If both of those look okay, then I can change this line in README.rst back to if cookiecutter.repo_hosting != "no".

> some badge urls hardcode "github", eg "https://landscape.io/github/blabla"

We could take the URL in three parts rather than one, but I think that would make things unnecessarily complicated.

I know nothing about how landscape.io works, but doesn't the fact that they refer to it solely as "github" rather than the full domain suggest that they do not, in fact, support other hosts?

It appears that requires.io only supports github.com or bitbucket.org. They connect directly, or something, and they won't connect to somebody's random GitLab server somewhere. The point is, if the service has an idiosyncrasy in what repo hosts it will work with, there's a certain unavoidable idiosyncrasy in how we can link to it. In the case of requires.io, we can put in special if-checks that put /github/ for github.com, /bitbucket/ for bitbucket.org, and nothing at all for any other host.
That makes an ugly little snarl of code around requires.io, but that makes a certain amount of sense to me, because requires.io *is in fact the source of the complication*.
If nicknaming github.com as "github" is common enough among other services, it might be worth making a little mapping of domain names to common_nicknames, since that is in fact what is actually happening. I can't immediately think of how to reference that dict in a jinja2 template, but I'm sure there must be a way.

Basically, if the service in question, e.g. landscape.io, treats github.com differently from any other host, then sooner or later *we* are going to have to treat gihub.com differently from any other host. The only question is whether we do that right at the point of the link to landscape.io, or at some further-away point where it's less clear why we're doing that.

> "repo_hosting": ["github.com", "gitlab.com", "custom domain", "private"],
> "repo_hosting_custom": 

...I mean, I don't see any particularly strong downside to that, except that it's a bit more complicated. But I don't see any upside at all. What specifically are you trying to prevent?

If we reference the `repo_hosting_custom` elsewhere, then the user still has the exact same free-text input to potentially screw up. So what's the point?
(If we wrap every single reference to `repo_hosting_custom` with an if-check on `repo_hosting`, we'll have a hideous snarl in every single link where we need to drop in the domain. I don't think that's what you were getting at?)

Certainly, no matter which way we do anything, the most common pathway for any user is going to be to hit enter without touching anything. And that ends up the same way whether it's a multiple-choice with a default or a text input with a default.

It would be nice if we could set it up so a user could type 1 for github.com, 2 for gitlab.com, or enter any custom URL and have it work. I do bemoan the loss of an in-your-face list of "these are *probably* the options you should be choosing from." But I don't think cookiecutter gives us that option, and we do always still have the default of github.com.

> for example I got a cffi binding for some crappy proprietary c stuff, I want to make sure it can never go on pypi by mistake.

Are you specifically bemoaning the loss of the obvious-in-your-faceness of the "no" option for repo_hosting, so that without the multiple-choice they have to check the README to know to type "no"?

I know nothing about what 'Private :: Do Not Upload' does in setup.py, but if it's important, it seems like we could do better by promoting that to its own standalone option. I could fold it into https://github.com/ionelmc/cookiecutter-pylibrary/pull/135, for example.

ionelmc: The 'Private :: Do Not Upload' is a classifier that makes pypi refuse the upload (special metadata) - https://twitter.com/di_codes/status/1097220464445931525 - I don't think it should be a separate option as it's rather a best practice thing.

Yes I am bemoaning the loss of the choice input. I forget things and I'd be redfaced if I have to check my own docs when I could have a god damn choice input :). As a matter of fact I do have to check my own docs on pytest-benchmark way too often. Plus I like to get a project by only using numpad.

Accept to have two option working in tandem? repo_hosting choices that solves my needs + repo_hosting_domain text input that solves your needs.

dHannasch: > The 'Private :: Do Not Upload' is a classifier that makes pypi refuse the upload (special metadata) - https://twitter.com/di_codes/status/1097220464445931525

Nifty!

I don't understand what that has to do with repo_hosting, though. I mean, you could have a private repo hosted on GitHub.com. You could have a private repo hosted on GitLab.com. You could have a package that *is* indexed on PyPI with the source repo being hosted by SourceForge. Why would you want to conflate indexing on PyPI with where the repository is hosted? That seems like exactly the kind of thing that means people need to read the docs very closely. The cause-and-effect there isn't exactly intuitive.

> I like to get a project by only using numpad.

I definitely see the advantages of not having to type anything, but do remember that that will always be the case for github.com, since that's the default. The only question is how much we want to do for gitlab.com.

In any case, I don't really see any downside to offering a choice upfront. That's a clever little way of working with what cookiecutter gives you. It's too bad we still have to prompt even if the user picks one of the standard choices, but c'est la vie. What do you think of https://github.com/ionelmc/cookiecutter-pylibrary/pull/129/commits/cded2ac8b5897b8d2370f145a505d2f295291773?

ionelmc: > Why would you want to conflate indexing on PyPI with where the repository is hosted? That seems like exactly the kind of thing that means people need to read the docs very closely. The cause-and-effect there isn't exactly intuitive.

> I added 7172421 to #135 to handle the case when the user wants something to not be uploadable. How does that look?

Ok you convinced me. Maybe name it "disallow_pypi_upload" (or something like that) to make it clear about it's effect? cookiecutter.json also needs to have that new option.

> In any case, I don't really see any downside to offering a choice upfront. That's a clever little way of working with what cookiecutter gives you. It's too bad we still have to prompt even if the user picks one of the standard choices, but c'est la vie. What do you think of cded2ac?

It seems find although I don't get why you have "standard" there.

dHannasch: > Maybe name it "disallow_pypi_upload" (or something like that) to make it clear about it's effect? cookiecutter.json also needs to have that new option.

To be clear, do you mean I should split the "pypi" option into two choices, one for the links to PyPI, one for the do-not-upload classifier?

...I guess it's true that there really are two choices, not one. On the other hand, that means even more choices at the top level, and it seems like there might be some potential for confusion between the two related PyPI choices.

What about one choice with two facets, as in https://github.com/ionelmc/cookiecutter-pylibrary/pull/135/commits/df1d84c52634d4667c765d221b00fecfcb3e001f?

> I don't get why you have "standard" there.

"standard" probably isn't the best word. I want to communicate the idea that we're asking the same question in two parts, where the first part just offers the opportunity to pick from a standard list instead of typing something in.

We can't give them the same name, because under the hood cookiecutter would just keep the second one, like if you do
```
    "full_name": "Ionel Cristian Maries",
    "full_name": ["some choice", "other choice"],
```
the second appearance of the key "full_name" is what matters and cookiecutter ignores the first.

So I was thinking maybe we could have a suffix denoting this as an "auxiliary" choice. E.g. if the real variable name is "repo_hosting", then the initial choice is repo_hosting_choose_from_list. Or something. The choices in the list that can be selected with a single keystroke are the ones that we thought were common enough to promote into the list, so they're sort of "standard" in a sense.

What do you think of repo_hosting_choose_from_list?

ionelmc: Too wordy ... too much text to read. The template is already intimidating with it's two dozen options. I think repo_hosting and repo_hosting_domain are the most brief but still clear enough. Users can't be confused about the second option since it would have a sensible default value if user didn't pick any of the special options.

dHannasch: Do you like it now? I think I tracked down all uses of repo_hosting and changed them to repo_hosting_domain.

PR #130, opened 2019-08-30
Remove os.unlink(join('ci', 'appveyor-download.py'))

dHannasch:
https://github.com/ionelmc/cookiecutter-pylibrary/commit/c426c169d0bec5faa5dace5f2c59865595ca77d8 removed appveyor-download.py, but post_gen_project.py still tries to unlink the non-existing file. This causes the generation to crash if cookiecutter.appveyor == 'no'.


ionelmc:
Thanks!

PR #132, opened 2019-09-02
Do not mention Travis in CONTRIBUTING.rst if travis==no

dHannasch:


{{cookiecutter.repo_name}}/CONTRIBUTING.rst
```diff
@@ -68,15 +68,17 @@ If you need some code review or feedback while you're developing the code just m
 
 For merging, you should:
 
-1. Include passing tests (run ``tox``) [1]_.
+1. Include passing tests (run ``tox``){%- if cookiecutter.travis == 'yes' %} [1]_{% endif %}.
 2. Update documentation when there's new API, functionality etc.
 3. Add a note to ``CHANGELOG.rst`` about the changes.
 4. Add yourself to ``AUTHORS.rst``.
 
+{%- if cookiecutter.travis == 'yes' %}
```
ionelmc: Note that the `-` will eat all the spaces so the result might not be valid rst.

dHannasch: Ah, good point. I think I want the `-` at the end of the if and the beginning of the endif. (Uh, I mean...see https://github.com/ionelmc/cookiecutter-pylibrary/pull/132/commits/eca7e204450944cb1810e10e6ae5efa7cd296b2f.) That way when cookiecutter.travis == 'yes' there are no extra spaces. (If cookiecutter.travis == 'no' there are a couple extra spaces between the end of this section and the beginning of the next section, but a couple extra spaces for a section break doesn't look so strange.)

Issue #93, opened 2018-05-21
Support for GitHub enterprise URLs


SterlingPeet:
Great project!

I am using this cookiecutter for some projects at Georgia Tech, most of which may eventually migrate to GitHub proper.  In the meantime, I have to go through a laborious process to change all the links to point at the enterprise url.  In this particular case, github.com gets changed to github.gatech.edu to convert over.  Is this something would be easy to change?

I don't have any experience writing cookiecutter templates, but I suppose I could try.

ionelmc:
I suppose we could have a `github_domain` variable in cookiecutter.json. Give it a shot if you want.

SterlingPeet:
Seems like this has been addressed 😄

PR #135, opened 2019-09-03
Add option for whether package is already on PyPI.

dHannasch:
I believe this will resolve the last little bit of #93.


dHannasch:
I just generated a package, with everything left as default except for the name:
```
$ ~/anaconda3/envs/python37env/bin/cookiecutter cookiecutter-pylibrary/ --overwrite-if-exists
project_name [Python Nameless]: Some Project Name
repo_name [python-nameless]: some-project-name
package_name [nameless]: some_project_name
distribution_name [python-nameless]: some-project-name
    To get started run these:
        cd some-project-name
        git init
        git add --all
```
Now, if I don't initialize git like the cookiecutter told me to, then tox -e check will get mad, so first I copy and paste from the instructions above:
```
/tmp/some-project-name$ git init
Initialized empty Git repository in /tmp/some-project-name/.git/
/tmp/some-project-name$ git add --all
```
Now then, tox.
```
/tmp/some-project-name$ ~/anaconda3/envs/python37env/bin/tox
...
writing output... [ 33%] contributing                                                                                            
(line   52) ok        https://tox.readthedocs.io/en/latest/install.html
writing output... [ 44%] index                                                                                                   
writing output... [ 55%] installation                                                                                            
writing output... [ 66%] readme                                                                                                  
(line    3) broken    https://pypi.org/project/some-project-name - 404 Client Error: Not Found for url: https://pypi.org/project/some-project-name/
(line    3) broken    https://pypi.org/project/some-project-name - 404 Client Error: Not Found for url: https://pypi.org/project/some-project-name/
(line    3) broken    https://pypi.org/project/some-project-name - 404 Client Error: Not Found for url: https://pypi.org/project/some-project-name/
(line    3) broken    https://pypi.org/project/some-project-name - 404 Client Error: Not Found for url: https://pypi.org/project/some-project-name/
(line   11) ok        https://readthedocs.org/projects/some-project-name/badge/?style=flat
(line   11) broken    https://readthedocs.org/projects/some-project-name - 404 Client Error: Not Found for url: https://readthedocs.org/projects/some-project-name/
build finished with problems.
ERROR: InvocationError for command /tmp/some-project-name/.tox/docs/bin/sphinx-build -b linkcheck docs dist/docs (exited with code 1)
  clean: commands succeeded
  check: commands succeeded
ERROR:   docs: commands failed
  py27: commands succeeded
  py35: commands succeeded
  py37: commands succeeded
  report: commands succeeded
```
This is what I meant. I mean that tox is telling them their build is failing. In particular, if they're using CI, it'll be telling them there's a problem with their package right out of the gate.

Of course, the broken readthedocs link is fixed by asking the user to input a working documentation link as sphinx_docs_hosting.
Someone might not have documentation on readthedocs, but they darn well oughtta have documentation *somewhere*, so it's actually kind of reasonable to tell them their build is failing if their documentation is inaccessible.

So one way to fix this is to allow pointing to a devpi instance instead of pypi.org, with a variable like sphinx_docs_hosting. I don't know how realistic that is. I haven't really looked into devpi, when I have a non-indexed package I just tell people to pip install it directly. If devpi is actually super-easy to set up, then it might be reasonable to just ask for a working index link, which defaults to pypi.org.


> we could make it also run the check env I guess?

We could, but like I said, on balance I think we shouldn't. tox *should* pass when we start, and in an ideal world it will. But depending on their environment, it might fail for all kinds of reasons outside our control. If tox is failing, we should let the user try to deal with it, rather than assume the cookiecutting failed.

Put another way, if we run tox and observe that it failed, there's nothing helpful we can actually *do* in that case --- the only thing we could do is report failure and delete the newly-created package, which isn't helpful.


> `twine upload` would give a clear error about some invalid `Private :: Do Not Upload` classifier, users can grep for that in the project and remove it if their mind changed regarding the pypi policy.

You're obviously correct, but like I said:

> the fact that it wouldn't *actually* be a big problem is irrelevant because *they don't know that when running cookiecutter*, which is when they are choosing options.

Unless we name the option disable_pypi_upload_but_dont_worry_its_easy_to_change_this_later_so_just_disable_it_if_youre_not_already_on_pypi, it's not obvious when using the cookiecutter that disabling upload is a low-consequence choice that nobody needs to stress over.

> You're suggesting you'll never use that upload breaking classifier?

I'll certainly find a use for it *sometimes*, and I'm sure you will too --- you mentioned having some ancient proprietary CFFI code that you need to be sure never gets exposed. https://github.com/ionelmc/cookiecutter-pylibrary/pull/129#discussion_r320426439
And I have a convenience package that includes a .ini file as a resource, not credentials or anything *really* sensitive, but a server address, a port number, a database name, things where I don't know exactly what harm they could do in the wrong hands but I'd really rather make sure they didn't get exposed.
I mean, I haven't actually bothered to disable uploading yet, simply because I cannot imagine any chain of events that would lead to me typing `twine upload` and the package name. If and when I ever get a devpi instance set up, then things will be different --- I'll be twine uploading all the time, and I'll want to make extra-double-sure that I don't upload something to PyPI when I meant to upload it to devpi, so I'll research all the layers of defense-in-depth I can pile on to prevent that.

But the point I was trying to make is that, *even among packages that are not (yet) on PyPI*, it'll probably be a minority that wants or needs to disable uploading. (Part of the "wants" will be just FUD about the disabling as I mentioned above, but I also separately think there's little point in trying to educate people on that point, because unless someone is regularly uploading to devpi, it's just not a feature *important* enough to be worth dragging people's attention to.) Not a tiny minority, but a minority. Whereas the population for whom the PyPI links will be broken on package creation will be...pretty much everybody.

ionelmc:
Merged with few fixes (ecccdcf). Unfortunately the dict as option triggers some bugs in cookiecutter in certain scenarios (cookiepatcher and probably with user configs too).

dHannasch:
Yeah, I noticed that `--config-file` didn't seem to be working as I expected and opened an issue on cookiecutter: https://github.com/cookiecutter/cookiecutter/issues/1231
I hadn't tried cookiepatcher.

I'm not sure what's going wrong --- cookiecutter never properly processed nested YAML, but I would have thought straight dumping the JSON dict into a config-file would work just like how we can enter a dict on the command line.
@lbrack seems to have a fix that would allow nested YAML in config-files https://github.com/lbrack/cookiecutter/commit/c9eb982b258d82e9fa7bbdff94e42bbe0316b9ac, don't know whether that would also allow cookiepatcher to handle dictionary variables.

PR #139, opened 2019-09-14
Fix bootstrap file not found error virtualenv

### LINKED ISSUE #138: bootstrap fails to bootstrap if virtualenv is not already installed
Similar to https://github.com/ionelmc/cookiecutter-pylibrary/issues/121, `cookiecutter gh:ionelmc/cookiecutter-pylibrary` crashes if virtualenv is not available on the command line. That's because bootstrap.py checks for subprocess.CalledProcessError, but the error we actually get is a FileNotFoundError. subprocess.CalledProcessError is the error that Python gives if we call python -m with a package name that is not installed.

```
Making bootstrap env in: /tmp/prednet/.tox/bootstrap ...
Traceback (most recent call last):
  File "ci/bootstrap.py", line 35, in <module>
    subprocess.check_call(["virtualenv", env_path])
  File "/home/david/anaconda/envs/python36env/lib/python3.6/subprocess.py", line 286, in check_call
    retcode = call(*popenargs, **kwargs)
  File "/home/david/anaconda/envs/python36env/lib/python3.6/subprocess.py", line 267, in call
    with Popen(*popenargs, **kwargs) as p:
  File "/home/david/anaconda/envs/python36env/lib/python3.6/subprocess.py", line 709, in __init__
    restore_signals, start_new_session)
  File "/home/david/anaconda/envs/python36env/lib/python3.6/subprocess.py", line 1344, in _execute_child
    raise child_exception_type(errno_num, err_msg, err_filename)
FileNotFoundError: [Errno 2] No such file or directory: 'virtualenv': 'virtualenv'
```

Even if we do catch the FileNotFoundError and reach `subprocess.check_call([sys.executable, "-m", "virtualenv", env_path])`, many Python 3 installations may not have virtualenv installed.
```
/home/david/anaconda/envs/python36env/bin/python: No module named virtualenv
subprocess.CalledProcessError: Command '['/home/david/anaconda/envs/python36env/bin/python', '-m', 'virtualenv', '/tmp/prednet/.tox/bootstrap']' returned non-zero exit status 1.
```

Even if virtualenv is installed, bootstrap.py will only get re-executed with the newly-created-by-virtualenv bootstrap environment if the new Python executable was created with a copy rather than a symlink. On some platforms it will be a symlink, so the virtualenv will be created but then it will go unused.

dHannasch:
Resolves #138.

### DIFF
--- a/{{cookiecutter.repo_name}}/ci/bootstrap.py
+++ b/{{cookiecutter.repo_name}}/ci/bootstrap.py
@@ -31,10 +31,7 @@ if __name__ == "__main__":
         import subprocess
 
         print("Making bootstrap env in: {0} ...".format(env_path))
-        try:
-            subprocess.check_call(["virtualenv", env_path])
-        except subprocess.CalledProcessError:
-            subprocess.check_call([sys.executable, "-m", "virtualenv", env_path])
+        subprocess.check_call([sys.executable, "-m", "virtualenv", env_path])
         print("Installing `jinja2` into bootstrap environment...")
         subprocess.check_call([join(bin_path, "pip"), "install", "jinja2"])
     python_executable = join(bin_path, "python")
--- a/{{cookiecutter.repo_name}}/ci/bootstrap.py
+++ b/{{cookiecutter.repo_name}}/ci/bootstrap.py
@@ -12,13 +12,6 @@ from os.path import exists
 from os.path import join
 from os.path import normpath
 
-try:
-    from os.path import samefile
-except ImportError:
-    def samefile(a, b):
-        return normpath(abspath(a)) == normpath(abspath(b))
-
-
 if __name__ == "__main__":
     base_path = dirname(dirname(abspath(__file__)))
     print("Project path: {0}".format(base_path))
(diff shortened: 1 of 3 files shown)


ionelmc:
Does the removal of `samefile(python_executable, sys.executable)` check + re-exec fixes the infinite exec problem? Does that really happen in practice?

There could be a command line argument that is added on execution (to avoid infinite execution) to avoid proliferation of extra files (to keep things neat).

dHannasch:
> Does the removal of `samefile(python_executable, sys.executable)` check + re-exec fixes the infinite exec problem? Does that really happen in practice?

...I just assumed that you were worried about some possible case and that's why you added the `samefile` check in the first place? Was that not the reason for the `samefile` check?

(I'm not really sure how to parse your question...if we literally just removed the `samefile` check and re-executed unconditionally, then by definition infinite recursion would happen 100% of the time.)

We could have a simple `!=` in place of the `samefile`. I don't know whether that would be safe or not. I don't know all the weird things that could happen if someone's home directory was accessed via a symbolic link on a samba share or something. I assumed that was the reason you used `samefile` instead of `!=` in the first place. (And the problem then becomes that `samefile` is actually too conservative on some systems --- it'll correctly report that the Python executable in a virtualenv is a symbolic link to the main Python executable, and therefore think it can just keep executing since it's already in the virtualenv...and then crash.)

> There could be a command line argument that is added on execution (to avoid infinite execution) to avoid proliferation of extra files (to keep things neat).

It seems like what we really have here is two functions `create_virtualenv_if_it_does_not_exist()` and `this_must_be_run_by_the_virtualenv()`. We want to run the first function with whatever Python is available, and then run the second function with the virtualenv.

Currently, we also re-run `create_virtualenv_if_it_does_not_exist()` with the virtualenv, but that's misleading --- that always has no effect. (If it did have an effect, that would be a bug.) Actually, it's weirder than that. We essentially have a recursive call, and `this_must_be_run_by_the_virtualenv()` is only run when the recursion terminates.
```
def create_virtualenv_if_it_does_not_exist():
    # do stuff...
    if not samefile: create_virtualenv_if_it_does_not_exist()
    else: this_must_be_run_by_the_virtualenv()
```
But...there isn't actually any recursion in the logic. We only actually need to create the virtualenv once. (I mean...right? Am I missing something? Do we sometimes need to create nested virtualenvs-of-virtualenvs?)

I'm sure there exists a way to do the recursion safely (without accidentally terminating the recursion too early as `samefile` can), but it seems easier to just skip the recursion if we don't need it.

We could use a command line argument.
```
if __name__ == "__main__":
    argparse...
    if args.definitely_being_run_by_virtualenv:
        this_must_be_run_by_the_virtualenv()
    else:
        create_virtualenv_if_it_does_not_exist()
        os.execv(python_executable, [python_executable, __file__, '--definitely-being-run-by-virtualenv'])
```
Of course we could skip argparse and just make `this_must_be_run_by_the_virtualenv()` into an actual, non-metaphorical function.
```
if __name__ == "__main__":
    create_virtualenv_if_it_does_not_exist()
    os.execv(python_executable, [python_executable, '-c', "import bootstrap; bootstrap.this_must_be_run_by_the_virtualenv()"])
```

Of course the simplest method (from the standpoint of execution, not from the standpoint of maintaining the cookiecutter) would be to make a tiny auxiliary Python package that installs an executable script into the virtualenv. Then we could just `subprocess.check_call([join(bin_path, "pip"), "install", "jinja2", "other-package-that-renders-jinja2-templates"])` and bootstrap.py could render the templates exactly the same way it creates the virtualenv:
```
def create_virtualenv_if_it_does_not_exist():
    subprocess.check_call([sys.executable, "-m", "virtualenv", env_path])
    return envExecutable

if __name__ == "__main__":
    envExecutable = create_virtualenv_if_it_does_not_exist()
    subprocess.check_call([envExecutable, '-m', 'other_package_that_renders_jinja2_templates'])
```
For that matter...if I'm understanding correctly how this all works, we don't want the `bootstrap` env for itself, we only want it to execute `this_must_be_run_by_the_virtualenv()`. If it just so happened that the system Python already had jinja2, we'd be perfectly content with that. So we could just
```
if __name__ == "__main__":
    try:
        this_must_be_run_by_the_virtualenv() # though actually the name this_must_be_run_by_the_virtualenv no longer makes sense if in fact it need not be run by a virtualenv
    except ImportError:
        envExecutable = create_virtualenv()
        os.execv(python_executable, [envExecutable, __file__])
```
but I would worry about infinite recursion in that case, because in some obscure corner-case `create_virtualenv()` might not solve the problem. If something goes wrong, better to crash out than to go into infinite recursion.

ionelmc:
> And the problem then becomes that `samefile` is actually too conservative on some systems --- it'll correctly report that the Python executable in a virtualenv is a symbolic link to the main Python executable, and therefore think it can just keep executing since it's already in the virtualenv...and then crash.

This is what I wanted to know - where does that happen? Like what virtualenv version/os?

dHannasch:
```
$ python -m venv somevenv
$ cd somevenv/
$ python -c "import os.path, sys; print(os.path.samefile('bin/python', sys.executable))"
True
$ python --version
Python 3.7.4
```
This is on Ubuntu 16.04.1 and Anaconda; I observed the same thing on Ubuntu 18.04, though I don't have access to that machine at the moment.

PR #142, opened 2019-09-18
Fix the changelog header-length calculation

dHannasch:
to match the length of `datetime.date.today().strftime("%Y-%m-%d")`.

This is hideously ugly, but there seems to be no way to call `datetime.date.today().strftime("%Y-%m-%d")` from jinja2 itself, nor can a cookiecutter pre_gen_project hook change the `cookiecutter.release_date` argument. So the only alternative to get a matching underline would be for the post_gen_project hook that inserts the date also insert an appropriate length of "-". (Which...I guess that would also be an option, if you'd prefer that.)

### DIFF
--- a/hooks/post_gen_project.py
+++ b/hooks/post_gen_project.py
@@ -52,9 +52,6 @@ if __name__ == "__main__":
 
 
     today = datetime.date.today()
-    replace_contents('CHANGELOG.rst',
-                     'hooks/post_gen_project.py replace this with datetime.date.today().strftime("%Y-%m-%d")',
-                     datetime.date.today().strftime("%Y-%m-%d"))
     replace_contents(join('docs', 'conf.py'), '<YEAR>', today.strftime("%Y"))
     replace_contents('LICENSE', '<YEAR>', today.strftime("%Y"))
 
--- a/hooks/post_gen_project.py
+++ b/hooks/post_gen_project.py
@@ -52,7 +52,7 @@ if __name__ == "__main__":
 
 
     today = datetime.date.today()
-    replace_contents('CHANGELOG.rst', '<TODAY>', today.strftime("%Y-%m-%d"))
+    replace_contents('CHANGELOG.rst', 'datetime.date.today().strftime("%Y-%m-%d")', datetime.date.today().strftime("%Y-%m-%d"))
     replace_contents(join('docs', 'conf.py'), '<YEAR>', today.strftime("%Y"))
     replace_contents('LICENSE', '<YEAR>', today.strftime("%Y"))
 
--- a/hooks/post_gen_project.py
+++ b/hooks/post_gen_project.py
@@ -52,7 +52,9 @@ if __name__ == "__main__":
 
 
     today = datetime.date.today()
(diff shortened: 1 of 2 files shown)


ionelmc:
I think the post_gen replacing it would be better. I don't even understand how this would work ... you would have a marker in there that looks like an expression ... that gets later replaced.

dHannasch:
> I think the post_gen replacing it would be better. I don't even understand how this would work ... you would have a marker in there that looks like an expression ... that gets later replaced.

Eh? Sorry, I don't understand. You think it would be better to have post_gen_project.py insert extra `-`, okay, fair enough. But specifically, the thing you don't like is that there's a marker in CHANGELOG.rst that later gets magically replaced in a non-obvious way. That's definitely a problem, but doing everything with post_gen_project.py will mean *more* of that thing you don't like, not less.

Maybe we should take a step back. The underlying problem here is that jinja2 does not provide a way to insert the current date, and cookiecutter's pre_gen_project hook does not provide a way to actually change the cookiecutter arguments such as `cookiecutter.release_date`.

You were able to hack around that by inserting a magic string in CHANGELOG.rst when `cookiecutter.release_date=='today'` and using post_gen_project.py to replace that magic string with `datetime.date.today().strftime("%Y-%m-%d")`.

That works, but there are two problems. Most importantly, the length of that string is not accessible within jinja2. The ugly hack I used here is just hardcoding the length of the string in as a magic number. That's obviously not very elegant, and it has the potential to trip us up later.

Speaking of tripping us up later, an ancillary problem is that depending on what the magic string is, it can be very non-obvious what's happening. I went all through the jinja2 documentation looking for a way to get the length of the `<TODAY>` string before I found out that it isn't a jinja2 thing at all. So if we do use a magic string, I want to make sure it's impossible to look at CHANGELOG.rst and not know what's going on. Sounds like maybe it's still unclear, so I'm pushing another commit that makes the magic string literally `hooks/post_gen_project.py replace this with datetime.date.today().strftime("%Y-%m-%d")`. Hopefully, that makes it clear that the magic string is going to be replaced with `datetime.date.today().strftime("%Y-%m-%d")`.

Anyway, as far as using post_gen_project.py to insert the `-`s we need, we can certainly do that, and then we wouldn't need to hardcode the length. That does mean we'll need to use another magic string. I think the best magic string for post_gen_project.py to replace with `'-' * len(datetime.date.today().strftime("%Y-%m-%d"))` is `post_gen_project.py replace this with '-' * len(datetime.date.today().strftime("%Y-%m-%d"))`. From a theoretical perspective it's not any more elegant than a magic string that's simply `TODAY` or `-TODAY-` or anything else, but that way anyone looking at CHANGELOG.rst immediately knows that there's magic happening in post_gen_project.py and it's not just a jinja2 thing.

dHannasch:
OR! We could use jinja2 magic to set a variable equal to a block, and thereby not need to do anything in post_gen_project.py at all.

ionelmc:
Yep, this is perfect. When I made the `<BLABLA>` thing there wasn't any `{% now %}` function available.

Issue #146, opened 2019-09-26
ENV=pure-private fails on Travis


dHannasch:
```
running check
warning: Check: missing required meta-data: url
error: Please correct your package.
ERROR: InvocationError for command /home/travis/build/ionelmc/cookiecutter-pylibrary/python-nameless/.tox/check/bin/python setup.py check --strict --metadata --restructuredtext (exited with code 1)
```
https://docs.python.org/3/distutils/examples.html#checking-a-package

distutils is complaining because no url argument is provided. This is intended behavior in setup.py,
```
{%- if cookiecutter.repo_hosting_domain != "no" %}
    url='https://{{ cookiecutter.repo_hosting_domain }}/{{ cookiecutter.repo_username }}/{{ cookiecutter.repo_name }}',
{%- endif %}
```
but distutils obviously doesn't like it. I don't immediately see a way to tell distutils to let the URL thing slide without eliminating the check altogether.

PR #149, opened 2019-09-26
Define block setup_requires_interior

### LINKED ISSUE #147: CHANGELOG.rst is not found by setup.py when testing on Travis
```
    Complete output from command python setup.py egg_info:
    Traceback (most recent call last):
      File "<string>", line 1, in <module>
      File "/tmp/pip-req-build-dzixo6zi/setup.py", line 45, in <module>
        re.sub(':[a-z]+:`~?(.*?)`', r'``\1``', read('CHANGELOG.rst'))
      File "/tmp/pip-req-build-dzixo6zi/setup.py", line 23, in read
        encoding=kwargs.get('encoding', 'utf8')
    FileNotFoundError: [Errno 2] No such file or directory: '/tmp/pip-req-build-dzixo6zi/CHANGELOG.rst'
```
Despite CHANGELOG.rst being in MANIFEST.in and `include_package_data=True` in setup.py, it appears that CHANGELOG.rst is not being included.

This appears to be the same problem I ran into on a different CI runner https://github.com/ionelmc/cookiecutter-pylibrary/issues/25#issuecomment-527262806, now manifesting on Travis.

Interestingly, two out of three SCM builds fail, but the third (cython) build passes. Maybe having 'cython' in `setup_requires` somehow makes `include_package_data` actually work? I really don't know what's going on there.

Aha: I think it's because the `setup_requires` in setup.cfg is ignored when we also have `setup_requires` in setup.py. This happens with the Cython build too, but (I think?) setuptools_scm is a dependency of Cython, so in the Cython case it ends up transitively included anyway.

I'm sort of scratching my head as to why Cython is in setup_requires, actually.
https://cython.readthedocs.io/en/latest/src/userguide/source_files_and_compilation.html?highlight=setup_requires#distributing-cython-modules
It looks like you only need Cython in setup_requires if you use `cmdclass={'build_ext': Cython.Build.build_ext}`, whereas you seem to use `setuptools.command.build_ext`.

dHannasch:
…that includes both pytest-runner and setuptools_scm as appropriate.

Resolves #147.

Note that this still has two Travis failures on ENV=pure-private (due to #146) and ENV=matrix-cext (that will be fixed by #145).

### DIFF
--- a/{{cookiecutter.repo_name}}/pyproject.toml
+++ b/{{cookiecutter.repo_name}}/pyproject.toml
@@ -1,0 +1,16 @@
+[build-system]
+requires = [
+    "setuptools>=30.3.0",
+    "wheel",
+{%- if cookiecutter.test_runner == 'pytest' and cookiecutter.setup_py_uses_test_runner == 'yes' %}
+    "pytest-runner",
+{%- endif %}
+{%- if cookiecutter.setup_py_uses_setuptools_scm == 'yes' %}
+    "setuptools_scm>=3.3.1",
+{%- endif %}
+{%- if cookiecutter.c_extension_support == 'cython' %}
+    "cython",
+{%- elif cookiecutter.c_extension_support == 'cffi' %}
+    "cffi>=1.0.0",
+{%- endif %}
+]
--- a/{{cookiecutter.repo_name}}/setup.cfg
+++ b/{{cookiecutter.repo_name}}/setup.cfg
@@ -37,6 +37,10 @@ max_line_length = 140
 #
 # Unless you're in one of those situations, you can simply ignore this.
 tests_require = {{ cookiecutter.test_runner }}
+
+[aliases]
+# Alias `setup.py test` to `setup.py {{ cookiecutter.test_runner }}`
+test = {{ cookiecutter.test_runner }}
 {% endif %}
 {% if cookiecutter.test_runner == "pytest" -%}
 
@@ -75,9 +79,6 @@ addopts =
     --tb=short
 {%- if cookiecutter.allow_tests_inside_package == 'yes' %}
     --pyargs
-testpaths =
-    {{cookiecutter.package_name}}
-    tests/
 # The order of these options matters. testpaths comes after addopts so that
 # {{cookiecutter.package_name}} in testpaths is interpreted as
 # --pyargs {{cookiecutter.package_name}}.
@@ -93,6 +94,9 @@ testpaths =
 # you can save a few milliseconds on testing by telling pytest not to search
 # the src/ directory by removing
 # --pyargs and {{cookiecutter.package_name}} from the options here.
+testpaths =
+    {{cookiecutter.package_name}}
+    tests/
 {%- else %}
 testpaths =
     tests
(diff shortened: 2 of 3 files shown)


ionelmc:
Oooof ... good catch. I don't really remember why I put it in setup.cfg.

ionelmc:
@dHannasch I'm thinking now about moving that tests_require in setup.py as well - but with a briefer comment - what do you think?

dHannasch:
> @dHannasch I'm thinking now about moving that tests_require in setup.py as well - but with a briefer comment - what do you think?

Hrm. It looks like they generally recommend using setup.cfg when possible, or rather, they want to recommend that but are worried about older versions before setup.cfg was supported. https://github.com/pypa/packaging.python.org/issues/378
As long as we're using setup.cfg at all, it seems like it probably makes sense to prefer leaving settings in setup.cfg rather than setup.py.

(Of course setup_requires *can't* be in setup.cfg because setup.cfg isn't sophisticated enough to handle things like "require CFFI but only sometimes." But tests_require *can* be in setup.cfg.)

To be honest I'm not sure why they recommend that, but if I had to guess it's because setup.cfg is easier to read, because there's usually some "mess" in setup.py to handle all kinds of oddball problems.

ionelmc:
Seems they are still talking about it (https://github.com/pypa/sampleproject/pull/57 and the mostly empty https://github.com/pypa/sampleproject/blob/master/setup.cfg). The `license_files = LICENSE.txt` is strange and out of place. They added that redundant thing (LICENSE* is already included) but didn't make a move on the other metadata.

I wouldn't trust them on anything regarding best practices tho, packaging.python.org is more of a cookbook that tries to give some overview on packaging. 

Give then fact that duplicated metadata in setup.py and setup.cfg ain't merged I'm thinking it's a bad practice to spread it across the two files, and it's impossible to have everything in setup.cfg. Therefore it's logical to conclude that the best practice is ... the opposite of https://github.com/pypa/packaging.python.org/issues/378 :)

ionelmc:
I'll merge this for now to avoid brokenness and we deal with the tests_require later maybe.

dHannasch:
To clarify, you're talking about deleting setup.cfg entirely, right? Or maybe not.

I can see the logic of, when multiple places might hold settings, keeping the settings only in the highest-priority place. That applies to pretty much everything in setup.cfg, because setup.cfg is never the highest-priority place. E.g. pytest.ini overrides setup.cfg, and the settings are not merged. http://doc.pytest.org/en/latest/customize.html

I guess I had been wondering whether we might want separate files .flake8, pytest.ini, etc.

dHannasch:
Incidentally, is your computer clock two days slow? 
0bee7f32d7e876853cadf461a0872d2f4df0d3e5 says it's from two days ago.
https://travis-ci.org/ionelmc/cookiecutter-pylibrary/builds/590801237?utm_source=github_status&utm_medium=notification

ionelmc:
Oooof ... ntp service was hosed :\

ionelmc:
Well I put what I could in setup.cfg to avoid dozens of root-level files. Coveragerc wasn't moved there because pytest-cov doesn't work nice with non-default file (it's messy and can't be fixed). I was suggesting moving out just the tests_require. Maybe we could move out all to separate files - don't know what to say - it's a try it and see how it looks thing.

PR #150, opened 2019-09-27
Use a file:// URL when the repository is local on the filesystem

### LINKED ISSUE #146: ENV=pure-private fails on Travis

dHannasch:
(rather than hosted).

Resolves #146.

This has three Travis failures, but those will be fixed by #145 and #149.

### DIFF
--- a/{{cookiecutter.repo_name}}/setup.py
+++ b/{{cookiecutter.repo_name}}/setup.py
@@ -12,6 +12,7 @@ import re
 import sys
 {% endif -%}
 from glob import glob
+import os.path
 from os.path import basename
 from os.path import dirname
 from os.path import join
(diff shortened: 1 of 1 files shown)


ionelmc:
Seems like a fair compromise.

PR #152, opened 2019-09-28
Print tox version on Travis.

dHannasch:


.travis.yml
```diff
@@ -37,6 +37,7 @@ env:
 before_install:
   - python --version
   - virtualenv --version
+  - tox --version
```
ionelmc: It should be last in the install section (this would run old tox or fail).

PR #153, opened 2019-09-28
A smaller implementation of #139 with a single file.

ionelmc:
@dHannasch so this is what I had in mind. From what I understand you have extremely lean system with no tooling like virtualenv/tox available on $PATH - this should solve without having extra files around.

### DIFF
--- a/{{cookiecutter.repo_name}}/ci/bootstrap.py
+++ b/{{cookiecutter.repo_name}}/ci/bootstrap.py
@@ -16,37 +16,34 @@ def check_call(args):
     print("+", *args)
     subprocess.check_call(args)
 
-if __name__ == "__main__":
-    base_path = dirname(dirname(abspath(__file__)))
-    print("Project path: {0}".format(base_path))
+def exec_in_env():
+    env_path = join(base_path, ".tox", "bootstrap")
+    if sys.platform == "win32":
+        bin_path = join(env_path, "Scripts")
+    else:
+        bin_path = join(env_path, "bin")
+    if not exists(env_path):
+        import subprocess
 
-    if "--no-env" not in sys.argv[1:]:
-        env_path = join(base_path, ".tox", "bootstrap")
-        if sys.platform == "win32":
-            bin_path = join(env_path, "Scripts")
-        else:
-            bin_path = join(env_path, "bin")
-        if not exists(env_path):
-            import subprocess
-
-            print("Making bootstrap env in: {0} ...".format(env_path))
+        print("Making bootstrap env in: {0} ...".format(env_path))
+        try:
+            check_call([sys.executable, "-m", "venv", env_path])
+        except subprocess.CalledProcessError:
             try:
-                check_call([sys.executable, "-m", "venv", env_path])
+                check_call([sys.executable, "-m", "virtualenv", env_path])
             except subprocess.CalledProcessError:
-                try:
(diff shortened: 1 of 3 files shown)


dHannasch:
This is basically the same as the argparse mockup in https://github.com/ionelmc/cookiecutter-pylibrary/pull/139#issuecomment-534244647, right?
```
if __name__ == "__main__":
    argparse...
    if args.definitely_being_run_by_virtualenv:
        this_must_be_run_by_the_virtualenv()
    else:
        create_virtualenv_if_it_does_not_exist()
        os.execv(python_executable, [python_executable, __file__, '--definitely-being-run-by-virtualenv'])
```
(Except `--definitely-being-run-by-virtualenv` is renamed to `--no-env` and the if-else is implicit rather than explicit.)

I mean, it'll work. It still seems really weird to me to have a file that recursively calls itself when we explicitly do not want any actual recursion to ever happen. That seems like it's just asking for bizarre bugs down the line.

ionelmc:
Yes it's basically that mockup. I added the recursion/reexec thing into bootstrap.py just to be able to run with without tox. Why you ask? Because in some configuration tox.ini is generated and I really wanted to have an eas way to run bootstrap.py when tox.ini is completely broken.

dHannasch:
It's good that you don't depend on tox, the whole bootstrapping business is a great idea, but I don't see the connection to recursion at all. If you didn't use recursion --- if you just called `create_virtualenv_if_it_does_not_exist()`, then called `this_must_be_run_by_the_virtualenv()` with the created virtualenv --- you still wouldn't depend on tox.

ionelmc:
Yeah but then I have 1 extra file around.

dHannasch:
(shrug) You do in fact have two different things that you need done, so...
In any case, like I said in https://github.com/ionelmc/cookiecutter-pylibrary/pull/139#issuecomment-534244647, if it bothers you to split the two functions across two files, they can be in one file explicitly `def`ed:
```
if __name__ == "__main__":
    create_virtualenv_if_it_does_not_exist()
    os.execv(python_executable, [python_executable, '-c', "import bootstrap; bootstrap.this_must_be_run_by_the_virtualenv()"])
```
Or at least make the command-line arguments into clear disjoint cases:
```
if __name__ == "__main__":
    if len(sys.argv[1:]) == 1:
        if sys.argv[1] == 'fill-templates':
            this_must_be_run_by_the_virtualenv()
        elif sys.argv[1] == 'create-env-and-fill-templates':
            python_executable = create_virtualenv_if_it_does_not_exist()
            os.execv(python_executable, [python_executable, __file__, 'fill-templates'])
        else: raise Exception
    else: raise Exception
```
Or if the full behavior must run with no command-line argument, then at least an explicit else-block:
```
if __name__ == "__main__":
    if len(sys.argv[1:]) == 1 and sys.argv[1] == 'fill-templates':
            this_must_be_run_by_the_virtualenv()
    elif len(sys.argv[:1]) == 0:
            python_executable = create_virtualenv_if_it_does_not_exist()
            os.execv(python_executable, [python_executable, __file__, 'fill-templates'])
    else: raise Exception
```

ionelmc:
Mkay, fair points about the argv. I did some refactoring, how does it look?

dHannasch:
Nice! I think the crucial `if __name__ == "__main__"` block is much easier to follow.

PR #154, opened 2019-09-28
Run doctests on all python versions

dHannasch:
It seems like if we have doctests, they are part of the test suite and should probably be run with all Python versions used to run the rest of the test suite.

### DIFF
--- a/{{cookiecutter.repo_name}}/ci/templates/tox.ini
+++ b/{{cookiecutter.repo_name}}/ci/templates/tox.ini
@@ -39,6 +39,9 @@ deps =
     coverage
     {%- endif %}
 {%- endif %}
+{%- if cookiecutter.sphinx_doctest == "yes" %}
+    sphinx
+{%- endif %}
 commands =
 {%- if cookiecutter.c_extension_support != "no" %}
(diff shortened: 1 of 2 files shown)


ionelmc:
Ooof... so there I a reason I didn't do it like this from the beginning - doctests are very fussy and hard to write to run on many python versions. The way it works now is - you either run them in the docs env (the point of them being: avoid broken examples) or you run them via pytest (where you have more options for easier testing). I think that is fine and shouldn't change.

dHannasch:
> doctests are very fussy and hard to write to run on many python versions.

That does sound like a reason to treat doctests as "second-class" tests and not try to make them work on all versions.
I've never personally observed the fussiness --- are you talking about things like unicode strings appearing differently in Python 2 output? Would it work to have them run on all py3* envs?

ionelmc:
Behold: a doctest suite -  https://github.com/ionelmc/python-tblib/blob/master/README.rst

I wish I never done something like that, it's a pain in the ass - trying to figure out what to replace with `...` to make the damn thing pass in all pythons. That's why doctests should be "second class" - a mere "example not totally broken validation" - to discourage ppl from jumping into something like that.

Issue #104, opened 2018-12-08
README: running "twine register" is no longer necessary


janLuke:
As written here: https://github.com/pypa/twine#twine-register

> WARNING: The register command is no longer necessary if you are uploading to pypi.org. As such, it is no longer supported in Warehouse (the new PyPI software running on pypi.org). However, you may need this if you are using a different package index.

You may consider to update the README.

PR #157, opened 2019-10-10
removed twine register command

joaomcteixeira:
`twine register` is no longer required or supported for PyPI.

```
HTTPError: 410 Client Error: Project pre-registration is no longer required or supported, upload your files instead. for url: https://upload.pypi.org/legacy/
```

I could upload just by doing:
```
twine upload --skip-existing dist/*.whl dist/*.gz
```

## changes
* removed `twine register` from `README`

(i just noticed it was already reported as issue)
Solves #104


ionelmc:
Thanks!

Issue #12, opened 2015-06-04
spell testenv fails: "The 'enchant' C library was not found"


cmeeren:
`tox -e spell` fails with the following error: `Could not import extension sphinxcontrib.spelling (exception: The 'enchant' C library was not found. Please install it via your OS package manager, or use a pre-built binary wheel from PyPI.)`

Full log:

```
$ tox -e spell
spell create: C:\Users\Christer\code\python\latexdiffcite\.tox\spell
spell installdeps: -rC:\Users\Christer\code\python\latexdiffcite/docs/requirements.txt, sphinxcontrib-spelling, pyenchant
spell develop-inst: C:\Users\Christer\code\python\latexdiffcite
spell installed: alabaster==0.7.4,Babel==1.3,colorama==0.3.3,docutils==0.12,Jinja2==2.7.3,latexdiffcite==0.9.0,MarkupSafe==0.23,pockets==0.2.4,pyenchant==1.6.6,Pygments==2.0.2,pytz==2015.4,six==1.9.0,snowballstemmer==1.2.0,Sphinx==1.3.1,sphinx-rtd-theme==0.1.8,sphinxcontrib-napoleon==0.3.6,sphinxcontrib-spelling==2.1.1,wheel==0.24.0
spell runtests: PYTHONHASHSEED='242'
spell runtests: commands[0] | sphinx-build -b spelling docs dist/docs
Running Sphinx v1.3.1

Extension error:
Could not import extension sphinxcontrib.spelling (exception: The 'enchant' C library was not found. Please install it via your OS package manager, or use a pre-built binary wheel from PyPI.)
ERROR: InvocationError: 'C:\\Users\\Christer\\code\\python\\latexdiffcite\\.tox\\spell\\Scripts\\sphinx-build.EXE -b spelling docs dist/docs'
___________________________________ summary ___________________________________
ERROR:   spell: commands failed
```

Excerpt from my `tox.ini`:

```
[testenv:spell]
setenv =
    SPELLCHECK = 1
commands =
    sphinx-build -b spelling docs dist/docs
usedevelop = true
deps =
    -r{toxinidir}/docs/requirements.txt
    sphinxcontrib-spelling
    pyenchant
```

Using Windows 7 64 bit, launching `tox` from Python 3.4.

ionelmc:
Looks like there aren't 64bit wheels for windows: https://pypi.python.org/pypi/pyenchant/

[...]

PR #159, opened 2019-10-10
Adds spell check in Travis

joaomcteixeira:
To my understanding the current version does not perform spell check on the CI servers. I found it very useful in my projects.

**Additions:**
- Added `spell` env to tox loop
- Added spell check run on Travis.
  - running spellcheck on Travis we can avoid running it on AppVeyor, right?
  - relates to #12 

Any additional discussions welcomed :-)

### DIFF
--- a/{{cookiecutter.repo_name}}/ci/templates/.travis.yml
+++ b/{{cookiecutter.repo_name}}/ci/templates/.travis.yml
@@ -1,6 +1,10 @@
 language: python
 dist: xenial
 cache: false
+addons:
+  apt:
+    packages:
+      - libenchant-dev
 env:
   global:
     - LD_PRELOAD=/lib/x86_64-linux-gnu/libSegFault.so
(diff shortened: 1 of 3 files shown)


ionelmc:
So there was a reason spell ain't in the CI configuration: it's a huge hassle to get it to return exit code 0 (you usually end up with ridiculous whitelist). In practice what I do is run it once in a while locally to check that I don't do any gross typos and ignore the exit code.

joaomcteixeira:
Hi, thanks for the quick reply.
Interesting,
I am using it with that configuration I proposed here. So, `spell` just runs in Travis and for py37 (in my case). Every time it passes in my tox locally it passes in the server, so far for this week that I have been testing it intensively. In some of my test repositories I added this last line to the `conf.py` file:

```
if os.getenv('SPELLCHECK'):
    extensions += 'sphinxcontrib.spelling',
    spelling_show_suggestions = True
    spelling_lang = 'en_US'
    # https://sphinxcontrib-spelling.readthedocs.io/en/latest/customize.html#private-dictionaries
    spelling_word_list_filename = ['spelling_wordlist.txt']  # <- this one
```
But in my last experiments it seems no to be necessary.

Can't add more info to the cause. Let me know your last thoughts.

PR #163, opened 2019-10-11
Remove automodule when running spellcheck.

dHannasch:
Also changes the name of the environment variable so it'll be easier to see where it comes from next time someone looks at this code.

This is another way to silence that
```
WARNING: autodoc: failed to import module 'sampleproject'; the following exception was raised:
No module named 'sampleproject'
```

It is not necessary to do anything with the `testsetup` that imports the module, because unlike autodoc, doctest already automatically doesn't do anything if you don't ask for it; the `testsetup` becomes inert and therefore no import occurs.

### DIFF
--- a/{{cookiecutter.repo_name}}/docs/conf.py
+++ b/{{cookiecutter.repo_name}}/docs/conf.py
@@ -19,14 +19,9 @@ if os.getenv('RUNNING_TOX_TESTENV_SPELLCHECK'):
     extensions += 'sphinxcontrib.spelling',
     # If RUNNING_TOX_TESTENV_SPELLCHECK, then we do not want to run autodoc,
     # because there is no need to spellcheck the autodocs.
-    # https://www.sphinx-doc.org/en/master/extdev/appapi.html#sphinx-core-events
-    def source_read_handler(app, docname, source):
-        for i,s in enumerate(source):
-            source[i] = s.replace('''.. automodule:: {{ cookiecutter.package_name }}
-    :members:''', '')
-    def setup(app):
-        app.connect('source-read', source_read_handler)
-    # alternate way to transform: https://github.com/sympy/sphinx-math-dollar/blob/master/sphinx_math_dollar/extension.py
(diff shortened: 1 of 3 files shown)


{{cookiecutter.repo_name}}/docs/conf.py
```diff
@@ -15,8 +15,18 @@
...
 ]
-if os.getenv('SPELLCHECK'):
+if os.getenv('RUNNING_TOX_TESTENV_SPELLCHECK'):
     extensions += 'sphinxcontrib.spelling',
+    # If RUNNING_TOX_TESTENV_SPELLCHECK, then we do not want to run autodoc,
+    # because there is no need to spellcheck the autodocs.
+    # https://www.sphinx-doc.org/en/master/extdev/appapi.html#sphinx-core-events
+    def source_read_handler(app, docname, source):
+        for i,s in enumerate(source):
+            source[i] = s.replace('''.. automodule:: {{ cookiecutter.package_name }}
```
ionelmc: Seems unreliable. Regex?

ionelmc:
But having 'source-read' handler just to silence warnings seems wrong in two ways:
- it's just a harmless warning
- it's doubling down on something wrong, why run those extensions anyway

Anyway, my proposal is to just remove the spell env. I've just tried it now on hunter it's horrendously cumbersome and broken - I have to whitelist names and things like docstring, backend or subclass. It's pointless really! <=  @joaomcteixeira

joaomcteixeira:
>  I have to whitelist names and things like docstring, backend or subclass. It's pointless really!

That's true for sure. hehe, I have no objections, go head and remove spellcheck altogether, definitively seems better than going around with complicated fixes just to try to maintain something that, as you said, is harmless and unmaintained.

Truth is that once one understands tox workflow it is easy to [add additional checks](https://github.com/joaomcteixeira/python-bioplottemplates/blob/851b004565372ebce9b9de6d3ceca46b348ef1de/tox.ini#L61-L63), and the main template should have only the most general and consistent approaches.

PR #164, opened 2019-10-12
Remove the spell env

ionelmc:
- it's cumbersome as you have to whitelist almost all programming jargon and names
- it's mostly and useless appendix - doesn't run in CI
- it relies on the unmaintained pyenchant lib

@dHannasch @joaomcteixeira now only thing left is to recommend some alternatives to checking spelling besides pycharm/other-big-IDE.

Closes #163 #160 #159 #12.

### DIFF
--- a/CONTRIBUTING.rst
+++ b/CONTRIBUTING.rst
@@ -48,7 +48,7 @@ To set up `cookiecutter-pylibrary` for local development:
 
    Now you can make your changes locally.
 
-4. When you're done making changes, run all the checks, doc builder and spell checker with `tox <https://tox.readthedocs.io/en/latest/install.html>`_ one command::
+4. When you're done making changes run all the checks and docs builder with `tox <https://tox.readthedocs.io/en/latest/install.html>`_ one command::
 
     tox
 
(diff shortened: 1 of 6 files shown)


ionelmc:
So bottom line: do you think it's still worth having the sphinx spellchecker?

dHannasch:
I don't really care about `testenv:spell` one way or the other, but I feel the need to point out that it turns out suppressing the warning isn't nearly as cumbersome as that thing you saw that I first pushed. It's literally one line (except for the comments).
https://github.com/ionelmc/cookiecutter-pylibrary/pull/163/files

ionelmc:
@joaomcteixeira please tell me you don't really need the spell env :)

joaomcteixeira:
Hehehe, I don't really need it, Vim has very nice spellcheck :P. And as said in the other PR discussion, I can add it manually if I want. I raised this discussion because I couldn't understand why spellcheck was not more integrated in the CI and tox loop; but now  with all the discussions you provided I understand your decisions and the strategy for the project.

> now only thing left is to recommend some alternatives to checking spelling besides pycharm/other-big-IDE

I will keep an eye on this and let you know if I find something.

ionelmc:
@joaomcteixeira 3.4 is dead: https://www.python.org/downloads/release/python-3410/

Issue #176, opened 2019-12-16
Add read the docs configuration file


mfaafm:
Hi, 
there is a new way to specify configuration of read the docs build using their ``.readthedocs.yml`` configuration files. The advantages are:

- The build environment can be configured in the repo
- The user does not have to change the advanced settings in the UI, since the ``.readthedocs.yml`` overrides the UI settings. 

I suggest to add it to the cookiecutter template.
If I see it correctly there are no settings in the template to change the docs location or similar, so a very simple configuration file shouls suffice.

This is the ``.readthedocs.yml`` I use in my project, which used your cookiecutter:
```yaml
# .readthedocs.yml
# Read the Docs configuration file
# See https://docs.readthedocs.io/en/stable/config-file/v2.html for details
version: 2

sphinx:
  configuration: docs/conf.py

formats: all

python:
  version: 3.7
  install:
    - requirements: docs/requirements.txt
    - method: pip
      path: .
```

dHannasch:
Looks good! From https://docs.readthedocs.io/en/stable/config-file/v2.html, right? Why did you delete the comments above e.g. `formats: all`?

For cleanliness' sake, it might be a good idea to remove .readthedocs.yml if `sphinx_docs_hosting` is not readthedocs. One way to do that would be in https://github.com/ionelmc/cookiecutter-pylibrary/blob/master/hooks/post_gen_project.py, e.g.
```
{%- if 'readthedocs' not in cookiecutter.sphinx_docs_hosting %}
    os.unlink('.readthedocs.yml')
{% endif %}
```
Same way we remove the docs directory if the user does not use Sphinx for documentation.

mfaafm:
> Looks good! From https://docs.readthedocs.io/en/stable/config-file/v2.html, right? Why did you delete the comments above e.g. formats: all?
I removed them in my project. For the cookiecutter I would keep them in. Also, I would document the step where the package is installed in the build to enable the autodoc features.

Good suggestion to remove the file! 
What happens when ``cookiecutter.sphinx_docs == "no"``? Do we need to cover that as well or does it suffice when we check ``cookiecutter.sphinx_docs_hosting``?

I was thinking like this, to always remove it when no sphinx docs are requested:
```python
{% if cookiecutter.sphinx_docs == "no" %}
    shutil.rmtree('docs')
    os.unlink('.readthedocs.yml')
{%- else %}
[...]

PR #177, opened 2019-12-24
Add .readthedocs.yml config file to configure Read the Docs build.

mfaafm:
Adds a ``.readthedocs.yml`` config file, which configures the Read the Docs build.
In the docs build, the package is installed using pip so the autodoc feature works. 
Since this config file overrides the advanced settings as configured in the UI, the user does not need to activate the virtualenv setting anymore. 

The ``.readthedocs.yml`` is removed if ``cookiecutter.sphinx_docs == "no" `` or if ``"readthedocs" is not in cookiecutter.sphinx_docs_hosting``. 

Implements changes suggested by #176.


ionelmc:
Looks good.

PR #186, opened 2020-02-27
Fix syntax error in .readthedocs.yml

### LINKED ISSUE #185: .readthedocs.yml syntax error
A `.readthedocs.yml` file was added per #176. But it has a syntax error due
to incorrect indentation which doesn't match the example provided in that issue.

nealmcb:
Fixes #185 

This indentation changes fixes a Read the Docs yaml parsing failure,
which surfaces as this Error:

Problem in your project's configuration. Parse error in .readthedocs.yml: YAML:
mapping values are not allowed here in "<unicode string>"
line 26, column 13: path: .

See more before-and-after in my r2b2 project: https://github.com/gwexploratoryaudits/r2b2/pull/5

Discussion at https://stackoverflow.com/a/41236822/507544

### DIFF
--- a/{{cookiecutter.repo_name}}/.readthedocs.yml
+++ b/{{cookiecutter.repo_name}}/.readthedocs.yml
@@ -23,4 +23,4 @@ python:
     - requirements: docs/requirements.txt
     # Install package using pip for the build (needed for autodoc)
     - method: pip
-        path: .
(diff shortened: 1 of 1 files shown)


ionelmc:
Thanks, don't know how I missed that.

PR #89, opened 2017-12-17
Add GPLv2+ and GPLv3+ licenses

leezu:

PR #191, opened 2020-03-30
Add GNU LGPL licenses.

riccardomurri:
This commit adds support for GNU LGPL v3 (or later), or v2 (or later)
to the license options.

Note that GNU LGPLv3 is an add-on to the GNU GPLv3, so the LICENSE file
includes both licenses in this case.

GNU licensing guidelines (see: https://www.gnu.org/licenses/gpl-howto.html)
recommend placing a copyright and license notice at the top of every
file; to keep consistency with how the other licenses are implemented
in `cookiecutter-pylibrary`, we put such a notice at the top of the
`LICENSE` file only.


riccardomurri:
Note: this adds support for the GNU *Lesser* GPL, so it's not the same as #89

{{cookiecutter.repo_name}}/setup.py
```diff
@@ -103,7 +103,12 @@ def _unavailable(self, e):
         "BSD 3-Clause License": "BSD-3-Clause",
         "MIT license": "MIT",
         "ISC license": "ISC",
-        "Apache Software License 2.0": "Apache-2.0"}[cookiecutter.license]
+        "Apache Software License 2.0": "Apache-2.0",
+        "GNU Lesser General Public License v3 or later (LGPLv3+)": "GNU Lesser General Public License v3 or later (LGPLv3+)",
```
ionelmc: This mapping here exists to map the displayed license (what the user sees when cookiecutter is used) to a a spdx complying identifier. See https://spdx.org/licenses/

riccardomurri: Thanks for the explanation! Fixed in c783640

ionelmc: Also, forgot to mention, it needs to be spdx because tidelift and I suspect other systems too use that standard to get license metadata (as opposed to the license classifiers that are already standardized in python, sadly).

PR #196, opened 2020-06-15
Update .gitignore.

virogenesis:
Updated/merged .gitignore from https://github.com/github/gitignore/blob/master/Python.gitignore

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,49 +1,151 @@
+# Byte-compiled / optimized / DLL files
+__pycache__/
 *.py[cod]
+*$py.class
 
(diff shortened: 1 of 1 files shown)


ionelmc:
Note that the gitignore that gets in the built project is here: https://github.com/ionelmc/cookiecutter-pylibrary/blob/master/%7B%7Bcookiecutter.repo_name%7D%7D/.gitignore

ionelmc:
This updates the repo's gitignore, not the template's gitignore (what the template produces) - I don't think we need this.

PR #210, opened 2020-11-01
Use --sitepackages to accelerate when we can.

dHannasch:
We could actually be even more specific than this, and allow system site packages for every individual testenv except the PyPy ones, or specifying `--no-build-isolation` for PyPy, but I'm not immediately sure how and it likely wouldn't be worth the bother.


ionelmc:
Honestly I'd just give up on system-site-packages. It's just asking for trouble for little gain. Moving to lxd builds already speed up the builds by a ton.

PR #211, opened 2020-11-03
Restrict some editorconfig settings to source files just in case

dHannasch:
...is what this turned into, but what it was really supposed to be was just adding a comment to explain a bit of what's going on in .editorconfig.

### DIFF
--- a/{{cookiecutter.repo_name}}/.editorconfig
+++ b/{{cookiecutter.repo_name}}/.editorconfig
@@ -10,7 +10,7 @@ indent_style = space
 # Things like specifying a charset will almost never cause problems,
 # but might cause mysterious problems in obscure cases,
 # so that's restricted to source-code files.
-[*.{py,pyx,pxd}]
+[*.{py,pyx,pxd,html,css,js,ts,c,cc,cpp,cxx,h,hpp}]
 insert_final_newline = true
 indent_size = 4
 charset = utf-8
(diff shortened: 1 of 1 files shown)


ionelmc:
I guess you have bumpversion in mind messing with that last line in the .ini file right?

ionelmc:
Tho I would rather add exceptions. I mean you add a html/css/js file in the django app and these settings won't pick up cause you forgot about this file.

dHannasch:
> you add a html/css/js file in the django app and these settings won't pick up cause you forgot about this file.

Hmm, that's a good point.

I'm not sure. It basically comes down to: when we stumble across a file we *didn't* anticipate, should we assume it's probably something like a source code file and apply the settings, or should we assume it's some mysterious unknown formatted-text resource file and not mess with it?

You're right that we know HTML is always UTF-8 and so forth, so we can definitely add HTML, and .c and .css and .js and .ts.

I get nervous about applying settings to unknown files. Like, if you were helping out with somebody else's code in person, and you saw a file with the extension .plg (for example) containing (say) a filepath and no trailing newline, would you say "Well, I have no idea what this is or what it's used for, but I'm going to go ahead and put a trailing newline here"?

(Sure, usually if people religiously git diff after any edit they'll catch any unexpected changes, but still.)

ionelmc:
Basically I would rather compromise on doing more than less. Eg: same with packaging - everything in src is grafted in the distribution, so you don't have to mess with the manifest for every new file type you include in the project. I think it's a good approach.

dHannasch:
(shrug) Fair 'nuff.

Issue #214, opened 2020-12-22
Shortcut option to disable python 2 versions


Querela:
As support for python 2 runs out, maybe it would be nice to have an option to disable all python 2 versions (python2.7, pypy) in the tox environments and the `setup.py`?

Alternatively, but it requires more work for postprocessing, an option to specify start and end python versions, e.g. 3.6 to 3.9.

It's more of an comfort feature as it's just deletion and modifications in various places. But it would make it easier and less error prone.
In my opinion, adding support for python2.7 for new packages should not be necessary anymore, and requirements for this should slowly vanish as time goes on. But completely removing it from the template might still be to soon. (Some edge cases might still exist.)

ionelmc:
Yes there could be an option for removing the py2 stuff. It could even be enabled by default (it's probably just me and few other people still maintaining py2 stuff). Are you interested in contributing?

Querela:
Well, that seems easy enough. So, yes, why not? :-)
Not sure when exactly I will get to it but before the year ends.
Any thoughts on the name of the feature? For cookie cutter?

ionelmc:
Well there are several options like "legacy_python" "disable_python2" "python2" "python3_only" or even "no_python2" - if you contribute you have freedom to choose whatever you like but I'd stay clear of double negations - it grinds my gears to see things like "no_something='no'".

Querela:
Ok. Makes sense. And yes, I wanted to stay away from double negations.
I was thinking about `python2` or with `_enable`/`_support`. Or `python3_only` but when python4 comes out, this will be confusing.
And without documentation, some variants are not really expressive when using cookiecutter ...
I think `legacy_python` might be good, as it describes all python versions that might not be supported anymore.
Alternatively, (future-proofing) `python2` [yes|no] which also allows for `python3` [yes|no] and `python4` [yes|no]

Querela:
Hmm.
I also updated the c extension code to remove the # macros for version checks.

In the appveyor config template I found no docs jobs? (just after checks)

Querela:
Should I update the Changelog file? There is no section for WIP/next version.

---

And I'm slightly confused about the `setup.cfg` for the tests:
```ini
legacy_python =
    : yes
    nolegacy: no
```
That's probably ok. But then I'm not how to reduce the number tests? (they doubled more or less) The `matrix` package did not really explain how to do this. And experimenting seems to work somewhat but is not really clear for me.
[...]

PR #216, opened 2020-12-22
Issue 214 legacy python2

Querela:
Addresses #214 

- added `legacy_python [no|yes]` cookiecutter switch
- updates `setup.py`, c extension code, `tox.ini`
  - set python2.7 as legacy
  - remove python3.5 completely (users can add it themselves later on)
- removes old `from __future__ import ...` if python2 is disabled, from `setup.py` only - `ci/bootstrap.py` can still be called with python2 even if project targets python3 only
- added tests for cookiecutter (WIP? - should they be reduced to improve test run time)
- add/update documentation to readme

_And I agree with the LICENSE the project is under._

### DIFF
--- a/{{cookiecutter.repo_name}}/setup.py
+++ b/{{cookiecutter.repo_name}}/setup.py
@@ -178,9 +178,6 @@ setup(
 {%- if cookiecutter.legacy_python == "no" %}
         'Programming Language :: Python :: 3 :: Only',
 {%- endif %}
-{%- if cookiecutter.legacy_python == "yes" %}
-        'Programming Language :: Python :: 3.5',
-{%- endif %}
         'Programming Language :: Python :: 3.6',
         'Programming Language :: Python :: 3.7',
         'Programming Language :: Python :: 3.8',
@@ -211,7 +208,7 @@ setup(
         # eg: 'keyword1', 'keyword2', 'keyword3',
     ],
 {%- if cookiecutter.legacy_python == "yes" %}
-    python_requires='>=2.7, !=3.0.*, !=3.1.*, !=3.2.*, !=3.3.*, !=3.4.*',
+    python_requires='>=2.7, !=3.0.*, !=3.1.*, !=3.2.*, !=3.3.*, !=3.4.*, !=3.5.*',
 {%- else %}
     python_requires='>=3.6',
 {%- endif %}
--- a/{{cookiecutter.repo_name}}/setup.py
+++ b/{{cookiecutter.repo_name}}/setup.py
@@ -1,7 +1,9 @@
 #!/usr/bin/env python
 # -*- encoding: utf-8 -*-
+{%- if cookiecutter.legacy_python == "yes" %}
 from __future__ import absolute_import
 from __future__ import print_function
+{%- endif %}
 
 import io
 {%- if cookiecutter.c_extension_support in ['yes', 'cython'] %}
@@ -169,7 +171,9 @@ setup(
         'Operating System :: POSIX',
         'Operating System :: Microsoft :: Windows',
         'Programming Language :: Python',
+{%- if cookiecutter.legacy_python == "yes" %}
         'Programming Language :: Python :: 2.7',
+{%- endif %}
         'Programming Language :: Python :: 3',
         'Programming Language :: Python :: 3.5',
         'Programming Language :: Python :: 3.6',
@@ -201,7 +205,11 @@ setup(
     keywords=[
         # eg: 'keyword1', 'keyword2', 'keyword3',
     ],
+{%- if cookiecutter.legacy_python == "yes" %}
     python_requires='>=2.7, !=3.0.*, !=3.1.*, !=3.2.*, !=3.3.*, !=3.4.*',
+{%- else %}
+    python_requires='>=3.5',
+{%- endif %}
     install_requires=[
 {%- if cookiecutter.command_line_interface == 'click' %}
         'click',
--- a/{{cookiecutter.repo_name}}/setup.py
+++ b/{{cookiecutter.repo_name}}/setup.py
@@ -176,7 +176,7 @@ setup(
 {%- endif %}
         'Programming Language :: Python :: 3',
 {%- if cookiecutter.legacy_python == "no" %}
-        # 'Programming Language :: Python :: 3 :: Only',
+        'Programming Language :: Python :: 3 :: Only',
 {%- endif %}
 {%- if cookiecutter.legacy_python == "yes" %}
         'Programming Language :: Python :: 3.5',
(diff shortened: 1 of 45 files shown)


{{cookiecutter.repo_name}}/src/{{cookiecutter.package_name}}/{{cookiecutter.c_extension_module}}.c
```diff
@@ -28,11 +28,15 @@ static PyObject* {{ cookiecutter.c_extension_function }}(PyObject *self, PyObjec
     PyObject *kwargs;
     PyObject *result;
 
+{%- if cookiecutter.legacy_python == "yes" %}
     #if PY_MAJOR_VERSION < 3
       module = PyImport_ImportModule("__builtin__");
     #else
+{%- endif %}
```
ionelmc, line 35: We probably need to have some creative positioning of the endif tag to make the indentation appear "natural". The goal of the template is to produce well looking output.

{{cookiecutter.repo_name}}/setup.py
```diff
@@ -169,9 +171,16 @@ def read(*names, **kwargs):
         'Operating System :: POSIX',
         'Operating System :: Microsoft :: Windows',
         'Programming Language :: Python',
+{%- if cookiecutter.legacy_python == "yes" %}
         'Programming Language :: Python :: 2.7',
+{%- endif %}
         'Programming Language :: Python :: 3',
+{%- if cookiecutter.legacy_python == "no" %}
+        # 'Programming Language :: Python :: 3 :: Only',
```
ionelmc, line 179: We can uncomment this, it's not like the user could install it on python2 anyway with the python_requires='>=3.6' constraint.

ionelmc:
Looking good, just few nits to fix. I think we can also simply drop python3.5 from everywhere, I don't expect anyone to need it considering how easy is to upgrade a minor py3 version.

{{cookiecutter.repo_name}}/setup.py
Querela: I was thinking about Python 4.

_python_requires could also be `~=3.6` I think to indicate not Python 4. But I would have to look again to be sure._

ionelmc: Live in the now, not in the whatcouldbe I'd say :-) Make things consistent.

{{cookiecutter.repo_name}}/src/{{cookiecutter.package_name}}/{{cookiecutter.c_extension_module}}.c
Querela: I was trying. I may need to reevaluate the templates again.

{{cookiecutter.repo_name}}/setup.py
Querela: Ok.

{{cookiecutter.repo_name}}/src/{{cookiecutter.package_name}}/{{cookiecutter.c_extension_module}}.c
ionelmc: Try this:
```
    #else
  {% endif %}    module = PyImport_ImportModule("builtins");
```

Querela:
> 
> 
> Looking good, just few nits to fix. I think we can also simply drop python3.5 from everywhere, I don't expect anyone to need it considering how easy is to upgrade a minor py3 version.

Ahh. You mean if users choose `legacy_python` (Py2.7 + Py3.5 + newer)? Or is there some other location were I forgot to remove it when choosing only new Python >= 3.6?

{{cookiecutter.repo_name}}/src/{{cookiecutter.package_name}}/{{cookiecutter.c_extension_module}}.c
Querela: I'm almost there.
I found two variants of macro indentation...
- https://github.com/ionelmc/cookiecutter-pylibrary/blob/052b10d6b5ac4067b685b9e202b6ba330023ddcc/%7B%7Bcookiecutter.repo_name%7D%7D/src/%7B%7Bcookiecutter.package_name%7D%7D/%7B%7Bcookiecutter.c_extension_module%7D%7D.c#L31-L35
- https://github.com/ionelmc/cookiecutter-pylibrary/blob/052b10d6b5ac4067b685b9e202b6ba330023ddcc/%7B%7Bcookiecutter.repo_name%7D%7D/src/%7B%7Bcookiecutter.package_name%7D%7D/%7B%7Bcookiecutter.c_extension_module%7D%7D.c#L98-L102
Can I simplify it to the second one that is flush to the left? or should I indent the second one, so it is better foldable (in editors) inside the function?

ionelmc:
I mean 3.5 is not legacy. It's simply something nobody cares about. Even pypy3 is at least 3.6. We can just remove it completely.

{{cookiecutter.repo_name}}/src/{{cookiecutter.package_name}}/{{cookiecutter.c_extension_module}}.c
ionelmc: Ooof you're right. Lets keep it simple and not ident macros at all.

Querela: Ok

Querela:
> 
> 
> I mean 3.5 is not legacy. It's simply something nobody cares about. Even pypy3 is at least 3.6. We can just remove it completely.

Oh? I know of many older linux distros that only ship with python3.5 at most. That's not great and I often use Anaconda to get the latest version. But that might be an use-case.
Otherwise, many fancy features are introduced in python3.6 (typings I think), so yea.

Querela:
> Mkay then lets lower the python_requires to >=3.5 then.

I might have misunderstood something. I thought you wanted to remove python3.5 from the legacy versions?
Currently in this pull request it is like this:
- with legacy python: `>=2.7, !=3.0.*, !=3.1.*, !=3.2.*, !=3.3.*, !=3.4.*`
- only python3: `>=3.6`

If users want to create software for legacy python versions, like python2.7 they also can do so for python3.5 and upwards. Deleting 3 locations where python3.5/py35 is mentioned, should be easy with searching.
The other option (`legacy_python=="no"`), they only have the choice of non-depreciated/in-support python versions, currently python3.6 upwards.

ionelmc:
I mean if you think it's valuable to still keep 3.5 for a while we can do it but then if legacy_python=no still keep 3.5 around for simplicity sake (avoid all sorts of weird branches in templates).

But it's a library template after all, so users are expected to have 3.6+ pythons around ... so I'm not really sure if the debian stretch argument is really that solid :)

Querela:
Ahh, no, `legacy_python=="no"` (the default) starts with python3.6.

ionelmc:
Very well, merged now, thanks.

PR #229, opened 2022-08-08
Insert missing space in docstrings

fwitte:
The example python -m command is missing a space between the "-m" and the package name.


ionelmc:
Hello, no it is not ☺ Just try it, it works just like that.

fwitte:
Oh wow, I was never aware of it… thank you! :)

PR #230, opened 2022-09-27
Fix toxinidir handling

cbeiraod:
Fixed references to toxinidir. When the directory contains spaces, the check-manifest command would fail because the path would be incorrectly interpreted as multiple options due to the spaces.

### DIFF
--- a/{{cookiecutter.repo_name}}/tox.ini
+++ b/{{cookiecutter.repo_name}}/tox.ini
@@ -130,7 +130,7 @@ skip_install = true
 commands =
     python setup.py check --strict --metadata --restructuredtext
 {%- if cookiecutter.setup_py_uses_setuptools_scm == 'no' %}
-    check-manifest '{toxinidir}'
+    check-manifest .
 {%- endif %}
 {%- if cookiecutter.linter == "flake8" %}
     flake8
--- a/{{cookiecutter.repo_name}}/tox.ini
+++ b/{{cookiecutter.repo_name}}/tox.ini
@@ -73,7 +73,7 @@ basepython =
         {%- endif -%}
     }: {env:TOXPYTHON:python3}
 setenv =
-    PYTHONPATH='{toxinidir}/tests'
+    PYTHONPATH={toxinidir}/tests
     PYTHONUNBUFFERED=yes
 {%- if cookiecutter.c_extension_support in ['yes', 'cython'] %}
     cover: SETUPPY_EXT_COVERAGE=yes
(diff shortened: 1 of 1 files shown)


{{cookiecutter.repo_name}}/tox.ini
```diff
@@ -73,7 +73,7 @@ basepython =
         {%- endif -%}
     }: {env:TOXPYTHON:python3}
 setenv =
-    PYTHONPATH={toxinidir}/tests
+    PYTHONPATH='{toxinidir}/tests'
```
ionelmc, line 76: But this doesn't need to be quoted right?

cbeiraod:
I quoted this line just to be safe... there is no harm in having the
quotes, but I didn't test without.

ionelmc:
Try running `tox -- python -c 'import sys; print(sys.path)'`

Lemme know what you get.

cbeiraod:
That gives a lot of output... Is there something in particular you would want to see?

The specific output of the python command was:
`['', "/Users/cristovao/Documents/Projects/PPS/LGAD Scripts/LIP_PPS_Run_Manager/'/Users/cristovao/Documents/Projects/PPS/LGAD Scripts/LIP_PPS_Run_Manager/tests'", '/usr/local/Cellar/python@3.9/3.9.10/Frameworks/Python.framework/Versions/3.9/lib/python39.zip', '/usr/local/Cellar/python@3.9/3.9.10/Frameworks/Python.framework/Versions/3.9/lib/python3.9', '/usr/local/Cellar/python@3.9/3.9.10/Frameworks/Python.framework/Versions/3.9/lib/python3.9/lib-dynload', '/Users/cristovao/Documents/Projects/PPS/LGAD Scripts/LIP_PPS_Run_Manager/.tox/py39/lib/python3.9/site-packages']`

There does seem to be something weird in the second path...

I get a couple of errors further down, but these are from missing local versions of python (I only have 1 on my local machine, but have the other versions configured so that the CI runs them all), and two additional errors on the docs and report, which I guess is expected since the command is passing something specific to tox.

ionelmc:
Note that still can break if quotes are in the path. Also see https://github.com/tox-dev/tox/issues/924

Did you try this?
```
commands =
    check-manifest .
```

cbeiraod:
Hi,

The `check-manifest .` approach also works fine. Let me know which approach you prefer and I can fix the PR if needed.

ionelmc:
Let go with that instead, it's the least problematic.

PR #245, opened 2024-04-03
Add required package to the README

tomek7667:
I wanted to use the template on a fresh machine and had all the deps but still got error:
```
➜  cyberman cookiecutter gh:ionelmc/cookiecutter-pylibrary
Unable to load extension: No module named 'jinja2_time'
```

Solved it by just installing it, but good to have up-to-date docs


pelrun:
Actually it was a bug in *cookiecutter*'s documentation - the reference to jinja2_time isn't needed and should be deleted instead.

https://github.com/cookiecutter/cookiecutter/issues/1888

